PR #10, opened 2022-01-18
Feat/chips

piitaya:
Add weather and entity chip.

Capture d’écran 2022-01-19 à 09 49 06

I also adding the ActionHandler from home assistant to easily support `tap_action`, `double_tap_action` and `hold_action`

📝 Todo in a next PR :
- add `icon_map` and `icon_color_map`options to adapt `icon_color` and `icon` with state 
- create a template chip to create custom chip


antonio1475:
double_tap_action doesn't seem to work?

```
type: custom:mushroom-light-card
entity: light.ordenador
hide_state: true
tap_action:
  action: toggle
hold_action:
  action: more-info
double_tap_action:
  action: more-info
show_brightness_control: true
use_light_color: true
show_color_temp_control: true
```
Neither on a chip card. Am I doing something wrong?
It would be great to have Double Tap Action on the Visual editor too!

piitaya:
We didn't include double tap but it's not a hard thing to add 🙂 I will create an issue.

PR #46, opened 2022-02-06
feat: add theme usage

piitaya:
The theme has moved to a now repository to be compatible with hacs : https://github.com/piitaya/lovelace-mushroom-themes


README.md
```diff
@@ -84,52 +84,7 @@ Some cards are on the todo list :
 
 ### Theme customization
 
-Mushroom works without theme but you can add a theme for better experience by copiing the theme file [themes/mushroom.yaml](themes/mushroom.yaml) in your theme folder.
-If you want more information about themes, check out the official [Home Assistant documentation about themes][home-assitant-theme-docs].
-
-```yaml
```
pierrecle, line 90: I think this should be in a dedicated doc file.

pierrecle (changes requested):
Very nice! I think the theme doc should be move to a dedicated file.
There is a more global point about the doc, maybe the wiki or a generated website, later, could be better than these files.

README.md
piitaya: The theme doc has moved to : https://github.com/piitaya/lovelace-mushroom-themes.
It's just a reference to this repo in a theme section.

pierrecle: I think it was too late for the review, I didn't saw the `-` at the beginning of the section...

pierrecle (approved):
👍

Issue #40, opened 2022-02-06
RGB color ramp ?


nightfever77:
Thanks for sharing your great work.
Would be great if you could add an RGB color ramp in addition to temperature and brightness in your future update.

thundergreen:
Also maybe fetch the effects if available? I am planning to make your light card my standard card :)

pierrecle:
> Also maybe fetch the effects if available? I am planning to make your light card my standard card :)

That’s a different thing, and it’s not planned for some reasons:
- we want to keep ui as light as possible
- as far as I know, effects are mainly used by automations, not manualy 
- we already have a good-but-not-great UX with the color selector, we have to enhance it before adding new elements in this card


Today you can create a grid with the light card and below another grid with buttons that call the right service.

thundergreen:
I agree maybe it would be overwhelming having to many options in the card . Thanks for the hint and keep up this great work .maybe for colour picker it could be something like settings button or expandable like with fold card ?

pierrecle:
Done in #48

Issue #35, opened 2022-02-05
Allow colored icon/slider corresponding to light color


pedolsky:
I'm deeply in love with your awesome project! 
Would it be possible to reflect the chosen light color in the icon and/or slider?

ImagineerNL:
+1

This assumes you can select a light color in this card? i was unable to do so. am i missing something or is this not possible?
(i can select the color in another card, but not throught the light card)

pedolsky:
Here is my code:

```
type: custom:mushroom-light-card
entity: light.hue_iris
vertical: true
show_brightness_control: true
show_color_temp_control: true
tap_action:
  action: toggle
hold_action:
  action: more-info

```
Have you defined the hold action?

ImagineerNL:
...i'm blind... totally forgot about the hold action.
Was figuring it would work the same as the temperature icon.

pierrecle:
I like the idea, I will take a look to color the icon and/or the slider.

pierrecle:
I'm not sure that coloring the slider is a good idea, it can make the whole component less understandable I my opinion.

pedolsky:
If you think it's going to be too confusing, better drop it. People love your project precisely because of its user-friendliness.

pierrecle:
https://user-images.githubusercontent.com/397503/152952452-bd5a0a53-d0e8-4b74-87aa-ad155587c818.MP4

When I look at the Apple expert UX designer solution, I guess I was wrong! I'll give a try today, it will be the last try before closing the first version of this feature 😉

pedolsky:
Love it! Thank you so so so much.

Only ```show_color_control``` doesn’t work, yet (maybe Apple related), but I didn’t expect a bonus :)

PR #48, opened 2022-02-06
feat(light): handle color

pierrecle:
Enregistrement de l’écran 2022-02-06 à 22 45 48

First proposition, solve #40 and related to #35 

Things to know:
- changing `color_temp` also change the RGB (HASS behaviour, cannot be changed)
- coldest `color_temp` is white, not blue
- for now, the only way I found to display lights colors in the right way is to set a dark background
- dark theme not tested

### DIFF
--- a/docs/cards/light.md
+++ b/docs/cards/light.md
@@ -11,16 +11,16 @@ A light card allow you to control a light entity.
 
 All the options are available in the lovelace editor but you can use `yaml` if you want.
 
-| Name                      | Type    | Default     | Description                                            |
-| :------------------------ | :------ | :---------- | :----------------------------------------------------- |
-| `entity`                  | string  | Required    | Light entity                                           |
-| `icon`                    | string  | Optional    | Custom icon                                            |
-| `name`                    | string  | Optional    | Custom name                                            |
-| `vertical`                | boolean | `false`     | Vertical layout                                        |
-| `hide_state`              | boolean | `false`     | Hide the entity state                                  |
-| `show_brightness_control` | boolean | `false`     | Show a slider to control brightness                    |
-| `show_color_temp_control` | boolean | `false`     | Show a slider to control temperature color             |
-| `show_color_control`      | boolean | `false`     | Show a slider to control RGB color                     |
-| `use_light_icon_color`    | boolean | `false`     | Colorize the icon according light temperature or color |
-| `tap_action`              | action  | `toggle`    | Home assistant action to perform on tap                |
-| `hold_action`             | action  | `more-info` | Home assistant action to perform on hold               |
+| Name                      | Type    | Default     | Description                                                       |
+| :------------------------ | :------ | :---------- | :---------------------------------------------------------------- |
+| `entity`                  | string  | Required    | Light entity                                                      |
+| `icon`                    | string  | Optional    | Custom icon                                                       |
+| `name`                    | string  | Optional    | Custom name                                                       |
+| `vertical`                | boolean | `false`     | Vertical layout                                                   |
+| `hide_state`              | boolean | `false`     | Hide the entity state                                             |
+| `show_brightness_control` | boolean | `false`     | Show a slider to control brightness                               |
+| `show_color_temp_control` | boolean | `false`     | Show a slider to control temperature color                        |
+| `show_color_control`      | boolean | `false`     | Show a slider to control RGB color                                |
+| `use_light_color`         | boolean | `false`     | Colorize the icon and slider according light temperature or color |
+| `tap_action`              | action  | `toggle`    | Home assistant action to perform on tap                           |
+| `hold_action`             | action  | `more-info` | Home assistant action to perform on hold                          |
--- a/docs/cards/light.md
+++ b/docs/cards/light.md
@@ -20,5 +20,6 @@ All the options are available in the lovelace editor but you can use `yaml` if y
 | `hide_state`              | boolean | `false`     | Hide the entity state                      |
 | `show_brightness_control` | boolean | `false`     | Show a slider to control brightness        |
 | `show_color_temp_control` | boolean | `false`     | Show a slider to control temperature color |
+| `show_color_control`      | boolean | `false`     | Show a slider to control RGB color         |
 | `tap_action`              | action  | `toggle`    | Home assistant action to perform on tap    |
 | `hold_action`             | action  | `more-info` | Home assistant action to perform on hold   |
--- a/docs/cards/light.md
+++ b/docs/cards/light.md
@@ -19,8 +19,8 @@ All the options are available in the lovelace editor but you can use `yaml` if y
 | `vertical`                | boolean | `false`     | Vertical layout                                        |
 | `hide_state`              | boolean | `false`     | Hide the entity state                                  |
 | `show_brightness_control` | boolean | `false`     | Show a slider to control brightness                    |
-| `use_light_icon_color`    | boolean | `false`     | Colorize the icon according light temperature or color |
 | `show_color_temp_control` | boolean | `false`     | Show a slider to control temperature color             |
 | `show_color_control`      | boolean | `false`     | Show a slider to control RGB color                     |
+| `use_light_icon_color`    | boolean | `false`     | Colorize the icon according light temperature or color |
 | `tap_action`              | action  | `toggle`    | Home assistant action to perform on tap                |
 | `hold_action`             | action  | `more-info` | Home assistant action to perform on hold               |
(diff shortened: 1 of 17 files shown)


src/cards/light-card/controls/light-color-control.ts
```diff
@@ -0,0 +1,110 @@
...
+        if (ctx) {
+            for(let i = 0; i < CANVAS_WIDTH; i++) {
+                let [r,g,b] = ctx.getImageData(i, 1, 1, 1).data;
+                if(r === rgb[0] && g === rgb[1] && b === rgb[2]) {
+                    return (i/CANVAS_WIDTH) * 100;
+                }
+            }
+        }
+        return 0;
+    }
```
piitaya, line 66: `_percentToRBG` percent is between 0 and 1
`_rgbToPercent` return number between 0 and 100

Maybe we can use the same range for the 2 functions?

The slider is a `hsv` color representation (with `s` and `v` fixed to 100%). 
We can use a `hsv` => `rgb` converter to avoid create dom element.

```ts
import * as Color from "color";

_percentToRGB(percent: number): number[] {
    const color = Color.hsv(360 * percent, 100, 100);
    return color.rgb().array();
}

_rgbToPercent(rgb: number[]): number {
    const color = Color.rgb(rgb);
    return color.hsv().hue() / 360;
}
```

This also fixed a bug: With canvas, if the `s` `v` of the provided color wasn't egal de 100% - 100% (by selecting a color with temp color slider for example), the percentage was wrong.

src/cards/light-card/light-card-config.ts
```diff
@@ -25,6 +25,7 @@ export const lightCardConfigStruct = assign(
         hide_state: optional(boolean()),
         show_brightness_control: optional(boolean()),
         show_color_temp_control: optional(boolean()),
+        show_color_control: optional(boolean()),
```
piitaya, line 28:
```suggestion
        show_color_control: optional(boolean()),
        use_light_icon_color: optional(boolean()),
```
I think we can split the 2 features for people who don't want colored icon by color control ?

src/cards/light-card/light-card-editor.ts
```diff
@@ -14,6 +14,10 @@ import { EditorTarget } from "../../utils/lovelace/editor/types";
 import { LIGHT_CARD_EDITOR_NAME, LIGHT_ENTITY_DOMAINS } from "./const";
 import { LightCardConfig, lightCardConfigStruct } from "./light-card-config";
 
+/*
+ * TODO: do not allow to set color and temp if bulb not compatible
+ */
+
```
piitaya, line 20: We can not disabled to field because the user can edit the value in yaml in yaml if he want. May be we can put a warning below or next the field ?

piitaya (commented):
Nice 🤗  ! I just suggested some improvements and fixes 🙂

piitaya:
For icon color, you can force the `saturationl` and `lightness` : no grey background need 🚀

https://user-images.githubusercontent.com/5878303/152757453-d61629e0-67e0-4ca0-b14e-ad096a191d65.mp4

```ts
const [r, g, b] = iconRgbColor
    .split(",")
    .map((c) => parseInt("" + c));
const color = Color.rgb([r, g, b]);

const transformedColor = color
    .hsl()
    .saturationl(100)
    .lightness(50)
    .rgb()
    .array();

let shapeColor = transformedColor;
let shapeAlpha = 0.2;
 ```

File: src/cards/light-card/controls/light-color-control.ts (hunk 1 above)
Comment (pierrecle) [reply src/cards/light-card/controls/light-color-control.ts:66]: Good point, too tired to have this idea, massive solution was easier at this point XD

**Comment by pierrecle** [comment]
> For icon color, you can force the `saturationl` and `lightness` : no grey background need 🚀 and it works on dark theme 🌓 Now we have the right color, maybe we can colorize the slider with this color too ?
> 
>  Enregistrement.de.l.ecran.2022-02-07.a.09.54.55.mp4 
> ```ts
> const [r, g, b] = iconRgbColor
>     .split(",")
>     .map((c) => parseInt("" + c));
> const color = Color.rgb([r, g, b]);
> 
> const transformedColor = color
>     .saturationl(100)
>     .lightness(50)
>     .rgb()
>     .array();
> 
> let shapeColor = transformedColor;
> let shapeAlpha = 0.2;
> ```

https://user-images.githubusercontent.com/397503/152882132-b70c3e1b-ed68-4729-8dce-d4926cd3cab9.mp4


I found that when setting `rgb_color`, `color_temp` is not set in the attributes so I could change the shapeColor only when changing `rgb_color`. The only problem is that with a particular temperature, the icon is not visible...

File: src/cards/light-card/light-card.ts [hunk 4]
```diff
@@ -142,6 +161,27 @@ export class LightCard extends LitElement implements LovelaceCard {
...
+                    .saturationl(100)
+                    .lightness(50)
+                    .fade(0.8)
+                    .rgb()
+                    .array();
+    
+                    iconStyle["--shape-color"] = `rgba(${transformedColor.join(',')})`;
+            }
+
+            iconStyle["--icon-color"] = `rgb(${iconRgbColor})`;
```
piitaya, line 205: Why one outside the condition and one inside ? It looks strange to have shape color different from icon color.

src/cards/light-card/utils.ts
```diff
@@ -12,6 +12,12 @@ export function getColorTemp(entity: HassEntity) {
         : undefined;
 }
 
+export function getRGBColor(entity: HassEntity) {
+    return entity.attributes.rgb_color != null
+        ? entity.attributes.rgb_color.join(',')
```
piitaya, line 17: Return the array of number to avoid split(',') and parseInt of the other side ?

src/cards/light-card/light-card.ts
pierrecle: Because the shape color is changed only when the rgb color is set, where the icon is always changed (color temperature or rgb).

piitaya: Yes but why? Change both or none ? The shape and the icon must have the same color to look better I think.

src/cards/light-card/utils.ts
pierrecle: Definitely better 😄

docs/cards/light.md
```diff
@@ -11,14 +11,16 @@ A light card allow you to control a light entity.
...
+| :------------------------ | :------ | :---------- | :----------------------------------------------------- |
+| `entity`                  | string  | Required    | Light entity                                           |
+| `icon`                    | string  | Optional    | Custom icon                                            |
+| `name`                    | string  | Optional    | Custom name                                            |
+| `vertical`                | boolean | `false`     | Vertical layout                                        |
+| `hide_state`              | boolean | `false`     | Hide the entity state                                  |
+| `show_brightness_control` | boolean | `false`     | Show a slider to control brightness                    |
+| `show_color_temp_control` | boolean | `false`     | Show a slider to control temperature color             |
+| `show_color_control`      | boolean | `false`     | Show a slider to control RGB color                     |
+| `use_light_icon_color`    | boolean | `false`     | Colorize the icon according light temperature or color |
```
piitaya, line 24: 👍😎

src/cards/light-card/light-card.ts
pierrecle: https://user-images.githubusercontent.com/397503/152889398-40c2423b-bd4e-4988-ab2d-16493bc93e28.mp4

Because with the coldest temperatures (whitish color), the background color don't have the same color (background turn to red and then yellow), and make it, in my opinion, quiet ugly and not truly understandable. Moreover, it's not that readable in coldest colors.

piitaya: And nobody select coldest color 😂 except if you like to live like in an hospital 🏨

pierrecle: We can be surprised, coldest light are used in bathrooms, kitchen, outside etc. And some lights are brighter on the coldest temp, you might use those temp in some occasion. I agree, that's not the temperature I use the most at home.
If the `mdi:lightbulb-on` had the same bulb size as `mdi:lightbulb` we could use a `blue-grey` background and the `mdi:lightbulb-on` to have a better contrast with `disabled` state and a more colored card by default.

piitaya: If it was up to me, I'll keep the base color if color temp is selected until we find a better solution. Do we just use the color of the light in rgb. Just apply custom color when rgb is active.
But if you prefer to go with the current solution, it's good to me too but it's an optional option 🙂

pierrecle:
https://user-images.githubusercontent.com/397503/152971476-89158a2b-050d-4905-b345-62f97185ac92.mp4


https://user-images.githubusercontent.com/397503/152971481-bb0e94c0-32af-4da6-b443-bba8b1e330d9.mp4

Inspired by Apple Home, finally this is, for me, the best solution 😃

Issue #51, opened 2022-02-07
[Feature Request] Possibility to wrap text overflow in template card.


DurgNomis-drol:
Love all these cards. The design and the simplicity is awesome!

I have a sensor that returns the pollen forecast as its state. This is a rather long text (200 characters)( I also believe that the max. character possible for a sensor in HA is 255). I would love if it was possible to wrap the text in the template card, either by default or as an option.

PR #59, opened 2022-02-07
Optional wrap secondary text in template card

DurgNomis-drol:
Adds the option to wrap text for the secondary row in the template card. 

Hope that PR's are welcome, though it is totally okay if you guys prefer not too! 😃

I am no frontend dev, nor TS developer, so please tell me if something can be done in a better.

I also don't speak French, so I used Google Translate for that one. 😄 

Fixes: #51 

image

### DIFF
--- a/src/translations/fr.json
+++ b/src/translations/fr.json
@@ -3,6 +3,7 @@
         "card": {
             "generic": {
                 "vertical": "Vertical ?",
+                "wrap_text": "Envelopper le texte?",
                 "hide_state": "Cacher l'état ?",
                 "state": "État",
                 "icon_color": "Couleur de l'icône",
--- a/src/translations/fr.json
+++ b/src/translations/fr.json
@@ -3,7 +3,7 @@
         "card": {
             "generic": {
                 "vertical": "Vertical ?",
-                "wrap_text": "Envelopper le texte?",
+                "multiline_secondary": "Secondaire multiligne?",
                 "hide_state": "Cacher l'état ?",
                 "state": "État",
                 "icon_color": "Couleur de l'icône",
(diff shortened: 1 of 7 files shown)


DurgNomis-drol:
Thanks!

That is a much better name 😃

src/translations/fr.json
```diff
@@ -3,6 +3,7 @@
         "card": {
             "generic": {
                 "vertical": "Vertical ?",
+                "multiline_secondary": "Secondaire multiligne?",
```
piitaya, line 6:
```suggestion
                "multiline_secondary": "Information secondaire sur plusieurs lignes?",
```
For the french translation 🙂

DurgNomis-drol: Awesome 🚀I hope my google translation did sound too weird 😆

piitaya:
👍 Can you update the documentation for this new option ? 🙂

Nesterovich2797:
How to wrap text in the right place yourself?

DurgNomis-drol:
I believe that the html tag works. Use `<br>`

Nesterovich2797:
image
I thought so too

piitaya:
I don't known if we can do that easily. But it's not a priority for now because it's an advanced usage and we are focused on more basic things for now. 

You can use [Home Assistant markdown card](https://www.home-assistant.io/lovelace/markdown/) for multiline texts.

You can also use [Mushroom title card](https://github.com/piitaya/lovelace-mushroom/blob/main/docs/cards/title.md) (it also does not support multiline).

image

DurgNomis-drol:
I actually forgot that it is not markdown where you can use HTML tags, but a Jinja2 template. But @piitaya is right, this is an advanced use case.

PR #60, opened 2022-02-07
Use relative time for entity card

DurgNomis-drol:
Display timestamp as relative time. This is also the default behavior in Home Assistant. We can also make this configurable if wanted.

### DIFF
--- a/src/cards/entity-card/utils.ts
+++ b/src/cards/entity-card/utils.ts
@@ -42,7 +42,7 @@ export function getInfo(
         case "name":
             return name;
         case "state":
-            if (entity.attributes.device_class === "timestamp") {
+            if (entity.attributes.device_class === "timestamp" && isAvailable(entity) == true) {
                 return html`
                 <ha-relative-time
                     .hass=${hass}
--- a/src/cards/entity-card/utils.ts
+++ b/src/cards/entity-card/utils.ts
@@ -42,7 +42,16 @@ export function getInfo(
         case "name":
             return name;
         case "state":
-            return state;
+            if (entity.attributes.device_class === "timestamp") {
+                return html`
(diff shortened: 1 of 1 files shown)


piitaya:
Hello 🙂
Thanks for your contribution !
Maybe we can have the same result with a template ?

DurgNomis-drol:
I did research that! BUT it only allows for timestamps in the past and not in the future as far as i can see https://www.home-assistant.io/docs/configuration/templating/#time

> `relative_time` converts datetime object to its human-friendly “age” string. The age can be in second, minute, hour, day, month or year (but only the biggest unit is considered, e.g., if it’s 2 days and 3 hours, “2 days” will be returned). Note that it only works for dates in the past.

DurgNomis-drol:
Also, displaying the relative time is the default behavior in the build-in cards.

piitaya:
I checked in HA code and they use this component.
```ts
<hui-timestamp-display
    .hass=${this.hass}
    .ts=${new Date(state)}
    format="relative"
    capitalize
></hui-timestamp-display>
```
Mabe you can try with this ?

piitaya:
Does `ha-relative-time` works with future date ?

DurgNomis-drol:
Yes it does 😃 There is also this option instead [hui-timestamp-display](https://github.com/home-assistant/frontend/blob/dev/src/panels/lovelace/components/hui-timestamp-display.ts) It have even more options if wanted. But as i know you guys wants to keep this simple (and I agree with this) I opted for the simple solution.

`hui-timestamp-display` gives a lot more options for how time is displayed.

piitaya:
Let's go for relative! We will add config options in it's a needed options for lot of people 🙂

PR #62, opened 2022-02-07
Align color and icon picker with other cards

DurgNomis-drol:
This adds the icon and color picker to this card as well, so there is feature parity between the card editors. 😃


piitaya:
It's by design. It's not an icon picker and a color picker because it's a template. You can use template in icon and icon_color, not just in primary and secondary.

DurgNomis-drol:
Right, that makes sense. Let's just closes this then 😊

Issue #47, opened 2022-02-06
[Request] Hide Icon


ahmadtawakol:
Can we have an option to hide the icons for all cards? Thanks, I'm loving these cards!

pierrecle:
Sorry but I can’t see the usecase..
It’s not planned, maybe in themes introduced by @piitaya you will be able to create one that hide the icons.

ahmadtawakol:
There are some entities that I wanted to display their state as text only, like the Thermal Comfort integration. I don't need an icon to be displayed, it just adds more clutter.

pierrecle:
Okay got it, I'll take it later ;)

pierrecle:
Fixed by #64

PR #64, opened 2022-02-08
feat(icon): allow user to hide icon in template and entity cards

pierrecle:
Capture d’écran 2022-02-08 à 01 39 18

Allow to hide icon on entity and template cards. In my opinion, not needed on every card.

Solves #47

### DIFF
--- a/src/cards/entity-card/entity-card.ts
+++ b/src/cards/entity-card/entity-card.ts
@@ -83,7 +83,6 @@ export class EntityCard extends LitElement implements LovelaceCard {
         const name = this._config.name ?? entity.attributes.friendly_name ?? "";
         const icon = this._config.icon ?? stateIcon(entity);
         const hideIcon = !!this._config.hide_icon;
-        console.log(hideIcon);
         const vertical = this._config.vertical;
 
         const stateDisplay = computeStateDisplay(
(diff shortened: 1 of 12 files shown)


src/cards/entity-card/entity-card.ts
```diff
@@ -122,13 +123,14 @@ export class EntityCard extends LitElement implements LovelaceCard {
                         hasHold: hasAction(this._config.hold_action),
                     })}
                     .hide_info=${primary == null && secondary == null}
+                    .hide_icon=${hideIcon}
                 >
-                    <mushroom-shape-icon
+                    ${!hideIcon ? html`<mushroom-shape-icon
```
piitaya, line 128:
```suggestion
                    ${!icon ? html`<mushroom-shape-icon
```
Maybe we can hide icon when no icon is defined in the template like for primary and secondary option 🙂

pierrecle: Yes, I was thinking about this. Maybe changing how `hideIcon` is initialiazed gives a better consistancy :)

src/cards/template-card/template-card.ts
```diff
@@ -114,6 +114,7 @@ export class TemplateCard extends LitElement implements LovelaceCard {
 
         const icon = this._templateResults.icon?.result;
         const iconColor = this._templateResults.icon_color?.result;
+        const hideIcon = !icon;
```
piitaya, line 117: So we can remove `hide_icon` option everywhere ?

pierrecle: > So we can remove `hide_icon` option everywhere ?

I did it in the last commit, for template only.

Issue #53, opened 2022-02-07
Add area Icon in person card


cyrilphoenix71:
Add the option to hide area name in the person card and put the area icon in tu upper left corner of picture (like suitcase for work area for my home)
With my previous issue we could have only picture and 2 icons

pierrecle:
Hide the state and the area name won't be displayed (neither the `Home` and `Away` information).
The badge where the icon is located won't be set on the left: in every application, the notification badge is set on the top right corner (maybe except `ltr` languages but I'm not sure) so no.

For now, we don't get the area icon, but this is something I'll do.

PR #65, opened 2022-02-08
feat(person): use zone icon when available

pierrecle:
https://user-images.githubusercontent.com/397503/152899442-abb962c1-7ce1-476c-b7b2-756899d1a093.mp4

We didn't use zone icon, that's done. There is no cleaner thing than friendly name, the zone entity id is not in the person attributes.

Solves #53

### DIFF
--- a/src/cards/person-card/utils.ts
+++ b/src/cards/person-card/utils.ts
@@ -7,7 +7,7 @@ export function getStateIcon(entity: HassEntity, zones:HassEntity[]) {
     } else if (state === "home") {
         return "mdi:home";
     }
-    const [zone] = zones.filter(z => state === z.attributes.friendly_name);
+    const zone = zones.find(z => state === z.attributes.friendly_name);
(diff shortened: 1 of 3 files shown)


src/cards/person-card/utils.ts
```diff
@@ -1,12 +1,16 @@
...
 
-export function getStateIcon(entity: HassEntity) {
+export function getStateIcon(entity: HassEntity, zones:HassEntity[]) {
     const state = entity.state;
     if (state === "unknown") {
         return "mdi:help";
     } else if (state === "home") {
         return "mdi:home";
     }
+    const [zone] = zones.filter(z => state === z.attributes.friendly_name);
```
piitaya, line 10:
```suggestion
    const zone = zones.find(z => state === z.attributes.friendly_name);
```
Better to use find 🙂

piitaya (approved):
Cool !

PR #66, opened 2022-02-08
fix(chip): vertically center weather icon and make icon-only chips round

pierrecle:
chip fix

On iOS, weather icon is not vertically align. Other thing fixed, icon-only chips are now a real circle.


piitaya (approved):
Nice ! I will edit the theme with these value when the theme PR will be merged.

PR #103, opened 2022-02-12
fix(cover): disable pause button if not in motion

brandensilva:
A small UX/UI improvement given that you cannot pause while in a stopped state.

**Examples:**

In a closed state pause is disabled:

image

While opening pause is available:

image

While closing pause is available:

image

In a opened state pause is disabled:

image


piitaya:
Some covers, like zwave ones, only reports their state at the end of the movement. Disable the button will remove the stop feature for this kind of cover.

brandensilva:
I see. Well the most important aspect I imagine is to maintain maximum compatibility with devices. Not sure there is really an easy solution unless you actually made it user controlled if they know what their device supports which doesn't seem nice to clutter the edit interface either. I'll go ahead and close this. I may merge it into in my fork still for my use case and keep the fork in sync with your upstream changes since I know my cover supports progression tracking.

PR #118, opened 2022-02-14
refactor: create mushroom-info-picker

piitaya:


pierrecle (approved):
Like this, I think we should start to split the editor, even if there is no shared component.

PR #123, opened 2022-02-14
feat(actions): add double tap action of all card

### LINKED ISSUE #122: Add double_tap action
Double tap is not supported and is a standard home assistant action : https://www.home-assistant.io/lovelace/actions/#double-tap-action

Add double tap support

piitaya:
fix #122


pierrecle (approved):
👌 
OK for now, but I think we'll have to make a MushroomCard class with common stuff and nearly all cards should extend this one. When we'll be able to handle tabs in configurator, we also should create common "global" configuration tab with hold, tab, double tap action etc.

PR #124, opened 2022-02-14
feat(chips): add alignment

### LINKED ISSUE #111: [Chips] Select chips horizontal alignment
User should be able to select the horizontal alignment of the chips.

Default would be as it is, `line start`, but user could select `line end`, `center` or `justify`

**Additional context**
The issue: https://github.com/piitaya/lovelace-mushroom-themes/issues/2

bemble:
Allow user to set alignment of chips.

closes #111

### DIFF
--- a/.hass_dev/views/chips-view.yaml
+++ b/.hass_dev/views/chips-view.yaml
@@ -162,3 +162,65 @@ cards:
               entity: light.bed_light
               tap_action:
                 action: toggle
+  - type: vertical-stack
+    title: Alignment
+    cards:
+      - type: custom:mushroom-chips-card
+        chips:
+          - type: weather
+            entity: weather.demo_weather_north
+          - type: weather
+            entity: weather.demo_weather_north
+            show_temperature: true
+          - type: weather
+            entity: weather.demo_weather_north
+            show_conditions: true
+          - type: weather
+            entity: weather.demo_weather_north
+            show_conditions: true
+            show_temperature: true
+      - type: custom:mushroom-chips-card
+        alignment: justify
+        chips:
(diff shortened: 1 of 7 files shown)


piitaya (approved):
Nice !
What about left, right instead of line-end line-start? 
We can create a alignment picker like the color picker too but it can be future refactoring 😁

pierrecle:
> Nice ! What about left, right instead of line-end line-start? We can create a alignment picker like the color picker too but it can be future refactoring 😁

Because `line-start` and `line-end` let us free about LTR and RTL stuff.
For the alignment picker, I thought about it and my conclusion is "if it's applicable for another component, sure we'll create an alignment picker".

piitaya:
Or just start and end like flex stuff ?

pierrecle:
> Or just start and end like flex stuff ? Or row instead of line ? I think "ligne" translation is "row" in this case.

`justify-content` uses `line-start` and `line-end`, that's why I used it :)

piitaya:
> > Or just start and end like flex stuff ? Or row instead of line ? I think "ligne" translation is "row" in this case.
> 
> `justify-content` uses `line-start` and `line-end`, that's why I used it :)

It's flex-start and flex-end 😅

pierrecle:
> > > Or just start and end like flex stuff ? Or row instead of line ? I think "ligne" translation is "row" in this case.
> > 
> > 
> > `justify-content` uses `line-start` and `line-end`, that's why I used it :)
> 
> It's flex-start and flex-end 😅

Hahaha, indeed, too tired :)
I change the translation and property values.

piitaya (approved):

PR #135, opened 2022-02-17
chore: add HA 2022.3 compatibility

### LINKED ISSUE #133: Move from paper drop down to mwc-select in editors
I would recommend moving away from Paper Drop down menus. I know this was not changed in this PR but something to think about. 

HA frontend is moving out all paper Drop downs and that means the component will no longer be loaded. We have moved to `mwc-select`

_Originally posted by @zsarnett in https://github.com/piitaya/lovelace-mushroom/pull/131#discussion_r809397488_

piitaya:
fix #133 

- [x] migrate color picker
- [x] migrate info picker
- [x] migrate layout picker
- [x] migrate fields in alarm editor
- [x] migrate fields in chips editor
- [x] migrate paper-input to mwc component
- [x] ~~use old dropdown for older home assistant version?~~
- [x] migrate mwc-textfield to mushroom-textfield
- [x] migrate mwc-textarea to mushroom-textarea 
- [x] migrate mwc-select to mushroom-select


piitaya:
@zsarnett I did some tests and the first HA version that introduce `mwc` components is the `2021.11` and luckily it's the same version for the icon picker. 🍀
We put `2021.11` as minimal version in the `hacs.json` file for the icon picker.
We can migrate all mushroom editors to the new `mwc` components! 🎉

zsarnett:
Bram is working on a better solution for custom cards now using Scoped REgistry. Thanks for the heads up on the mwc-textarea.

PR #169, opened 2022-02-27
feat(cover): adapt buttons with cover features

### LINKED ISSUE #149: Cover card: Pause button is displayed at all times even if entity don't support that
**Describe the bug**
The pause button is displayed at all times. It makes no sense to display it if the blind is not moving or the entity does not support SUPPORT_STOP. 

**Expected behavior**
Hide pause button when is not supported

**Describe alternatives you've considered**
Maybe it's a good idea to allow the user to use the buttons as a quick action. Maybe set the blinds to 50%. For example, I use three settings. Blinds fully open, blinds shaded, and blinds darkened. It can be handy for the user to be able to set the behavior of the buttons. 

**Screenshots**
image

### LINKED ISSUE #163: Cover Card buttons disabled if motors not report state
I have a problem with Cover Card.
I use Somfy RTS roller motors.
The problem with these roller motors is that they do not report the state.
They accept 4 commands:

UP
Down
Stop
Pairing
If i want to open 20% of the roller ( assuming it is closed) i press UP and after that Stop when I consider that I have reached the 20%.

My problem: when i press Up button the cover card switch state from Closed to Open and disable the button Up. After that if i want to open to 50% or 100% i have to press Down to change the state to Closed , then Stop and Up to open.
Every time I want to change roller position , I have to do something like that.
I was wondering if there was a possibility that the buttons would remain active regardless of roller position?

piitaya:
New rules :
- better disabled state for buttons
- disable buttons in entity is unavailable
- don't show pause, open and close buttons if not supported

fix #163 
fix #149

### DIFF
--- a/src/cards/cover-card/controls/cover-buttons-control.ts
+++ b/src/cards/cover-card/controls/cover-buttons-control.ts
@@ -4,6 +4,7 @@ import { css, CSSResultGroup, html, LitElement, TemplateResult } from "lit";
 import { customElement, property } from "lit/decorators.js";
 import { classMap } from "lit/directives/class-map.js";
 import "../../../shared/button";
+import { UNAVAILABLE } from "../../../utils/entity";
 import { computeCloseIcon, computeOpenIcon } from "../../../utils/icons/cover-icon";
 import {
     isClosing,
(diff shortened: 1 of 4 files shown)


src/cards/cover-card/controls/cover-buttons-control.ts
```diff
@@ -36,25 +44,63 @@ export class CoverButtonsControl extends LitElement {
         });
     }
 
+    private _computeOpenDisabled(): boolean {
+        if (this.entity.state === "unavailable") {
```
zsarnett, line 48: Creating an "unavailable" constant variable would be good for code quality

src/cards/cover-card/controls/cover-buttons-control.ts
```diff
@@ -36,25 +44,63 @@ export class CoverButtonsControl extends LitElement {
         });
     }
 
+    private _computeOpenDisabled(): boolean {
```
zsarnett, line 47: You could also just make these getters

zsarnett (approved):

PR #170, opened 2022-02-28
Feat/fan auto control

davidhazell:
I wanted the ability to set my Levoit and Dyson fans to auto speed so I added a button for this on the mushroom fan card.

### DIFF
--- a/docs/cards/fan.md
+++ b/docs/cards/fan.md
@@ -20,6 +20,7 @@ All the options are available in the lovelace editor but you can use `yaml` if y
 | `hide_state`              | boolean | `false`     | Hide the entity state                                                     |
 | `icon_animation`          | boolean | `false`     | Animate the icon when fan is `on`                                         |
 | `show_percentage_control` | boolean | `false`     | Show a slider to control speed                                            |
+| `show_auto_control`       | boolean | `false`     | Show a button to set auto speed                                           |
 | `show_oscillate_control`  | boolean | `false`     | Show a button to control oscillation                                      |
 | `tap_action`              | action  | `toggle`    | Home assistant action to perform on tap                                   |
 | `hold_action`             | action  | `more-info` | Home assistant action to perform on hold                                  |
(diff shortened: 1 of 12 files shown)


davidhazell:
Screenshot 2022-02-27 230911

piitaya (commented):
Thank's for your PR.
I feel like it's more preset mode than auto mode.
Your fan only support auto mode. Mine support two presets : 
- Nature
- Normal
- Maybe we can find a way for the user to select the mode ?

davidhazell:
I thought about doing that but thought the UI may get too busy for the minimalist style and figured auto was a common option like oscillate. I suppose I could do something similar to the light card 'temperature' and 'color' buttons that repurpose the slider area.

Maybe a "preset_mode" button that displays the available presets where the slider normally is?  My only concern is for fans with many presets available and fitting the buttons.

I could also make the 'auto' button a configurable 'preset_mode' button where the user defines a single preset mode to be set (auto, sleep, nature, etc) and allow the user to configure the icon.

What do you think?

piitaya:
Hi ! With @bemble, we are discussing about how we can create the climate card with `hvac_mode` and `preset_mode`. I think it can be a good thing to have something similar for the fan `preset_mode`.

davidhazell:
I agree. It might make sense to have a similar design for each card's `preset_mode` selection interface whether that's a dropdown selection, individual buttons, etc to keep the interface consistent. Have @bemble and you discussed how you'd like to display `preset_mode` selection?

mihailescu2m:
Any progress on this?
Using chips for presets?

robert4y:
Add the remaining time until the fan turns off when the timer is running

piitaya:
Thank for your contribution. I'm closing the PR as it can be achieve 

I'm closing this PR as it can be done using tile and [fan preset mode card feature](https://www.home-assistant.io/dashboards/features/#fan-preset-modes).
I do not prefer to duplicate the feature from HA in Mushroom. I'm looking to provide card features support to Mushroom.

PR #179, opened 2022-03-01
feat(template): add entity in template card and chip

piitaya:
It is now possible to use entity in template

```
{{ states(entity) }}
```

The `more-info` and `toggle` action use this entity too.


bemble (approved):
Ok for me, thanks!

zanna-37:
I think the documentation should be updated here https://github.com/piitaya/lovelace-mushroom/blob/main/docs/cards/template.md

PR #181, opened 2022-03-02
fix(ha): hack to load ha component

piitaya:
These components are not loader by default with 2022.3 version. 
This hack loads the components (temporary hack before an official HA solution)


src/cards/title-card/title-card-editor.ts
```diff
@@ -15,6 +16,10 @@ export class TitleCardEditor extends LitElement implements LovelaceCardEditor {
 
     @state() private _config?: TitleCardConfig;
 
+    protected firstUpdated(): void {
```
zsarnett, line 19: You can also do this on connectedCallback() if you find this not loading quickly enough

Issue #109, opened 2022-02-13
Select 'Attribute' as information source for entity card


Gurutasker:
Many sensors and entities store information as attributes. 
Often this is desired to be displayed on an entity card, or in some cases other cards instead of the 'state' of the entity.

For example, a sensor from a router integration may show a port status as 'up' and the attributes information such as IP Address, RX/TX information.

If attributes could be selected as an information source in the entity card this would allow for significantly more flexibility in the information shown. 

As an added feature and in keeping with the intent of the project, the attributes could be enumerated from the entity to be selectable in a drop-down list.

Entity Card

Primary information source 
  -> Attributes
     -> _Available attributes listed here_

cyrilphoenix71:
+100000

piitaya:
We already discussed about that https://github.com/piitaya/lovelace-mushroom/issues/45. We do not plan to add attribute  on entity card for now because this can be solve with template-card because it's an advanced usage.
Unlike the state, attributes can't be formatted and translated. If we add attribute, we will have a lot of issues : "add unit to attribute", "allow to select 2 attributes", etc... So we prefer for now to use template for that to add maximum flexibility.

Note : we will add `more-info` tap action to template card soon 🙂

```
type: custom:mushroom-entity-card
primary: My router
secondary: {{ state_attr('sensor.my_router', 'ip_address') }} 
icon: mdi:router
color: blue
```

luixal:
I was going to ask for something similar, instead of attributes, values from another entity. In my case, almost all of my switches have a power meter inside, which is configured in HA as **three** different entities:
 
![image](https://user-images.githubusercontent.com/615268/153905848-24394c45-4c07-4a53-933a-53f843ce1849.png)

It would be nice to be able to select the value of one of those entities here:

![image](https://user-images.githubusercontent.com/615268/153905792-d37063db-5e41-48fd-9141-918ca8a72e8e.png)

so I can show consumption value while turned on. I've managed to do it with an entity card. If someone else gets here looking for something like this, this is how I did it:

```yaml
- type: custom:mushroom-template-card
  entity: switch.lamapara_salon_mss310_main_channel
  icon: mdi:floor-lamp-outline
  icon_color: "{% if is_state('switch.lamapara_salon_mss310_main_channel', 'on') %} orange {% endif %}"
  primary: Lámpara
  secondary: >
[...]

PR #191, opened 2022-03-08
Feature/attribute card

tomhoogeveen:
Mushroom

Hi!
I just discovered your project and I like it a lot. But I was missing a way to show some more information(attributes) of an entity. So I just went tinkering around and made this Attribute card. 
It is based on the entity-card, so you might see some similarities.

Not a 100% happy with the design but it's functional.

I have no idea if you are at this point in time looking for anything like this but I thought might as well try :)

### DIFF
--- a/src/cards/attribute-card/attribute-card-config.ts
+++ b/src/cards/attribute-card/attribute-card-config.ts
@@ -14,8 +14,8 @@ export interface AttributeCardConfig extends LovelaceCardConfig {
     hide_icon?: boolean;
     layout?: Layout;
     primary_info?: Info;
-    secondary_info?: Info;
-    attribute?: Attributes;
+    primary_attribute?: Info;
+    secondary_attribute?: Attributes;
     tap_action?: ActionConfig;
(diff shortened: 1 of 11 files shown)


piitaya:
Hi 🙂 thanks for the PR. Why do not use the template card to do this? 

```yaml
type: custom:mushroom-template-card
entity: climate.thermostat
primary:  '{{ state_attr(entity, "name") }}'
secondary:  '{{ state_attr(entity, "température") }}°C'
```

piitaya:
What about this ?
```yaml
type: custom:mushroom-template-card
entity: climate.thermostat
primary:  '{{ state_attr(entity, "name") }}'
secondary:  '{{ state_attr(entity, "temperature") }}°C - {{ state_attr(entity, "battery") }}%'
```
The problem with an attribute card is that attributes are not normalized is Home Assistant so it can be difficult to have rules for everyone. Also, it can be difficult to format/translate.
That's why we recommend to use template card for this custom use-cases.
We already discussed about attributes in the issue #109.

PR #201, opened 2022-03-18
feat(theme): add font size/weight theme variable for cards and chips

### LINKED ISSUE #198: Allow change font size
Hi! 
Your work is very good, congratulations :)
I would like to be able to change the primary and secondary (name and status) font size on the card.
I tried to overwrite css in the mushroom.js file, but I didn't succeed. Even if I change it there, something always overwrites it. Have any idea? Unfortunately, the letter is very small for me (14-12px).
Have a nice day!

piitaya:
Fix #198


fantnhu:
thank you very much! wonderful! :)

PR #218, opened 2022-04-01
feat(slider): add single tap gesture to slider

Hooolm:
First of all, love your work!

I would suggest setting this threshold to 0, or make it configurable. This make the controls of lights,covers, colors etc more snappy. And you do not have to pull it to change it rather than tapping it to change it.

### DIFF
--- a/src/shared/slider.ts
+++ b/src/shared/slider.ts
@@ -63,11 +63,14 @@ export class SliderItem extends LitElement {
             this._mc = new Hammer.Manager(this.slider, { touchAction: "pan-y" });
             this._mc.add(
                 new Hammer.Pan({
-                    threshold: 0,
+                    threshold: 10,
                     direction: Hammer.DIRECTION_ALL,
                     enable: true,
                 })
             );
+
+            this._mc.add(new Hammer.Tap({ event: "singletap" }));
+
             let savedValue;
             this._mc.on("panstart", () => {
                 savedValue = this.value;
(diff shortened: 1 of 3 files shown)


Hooolm:
I found that this is in fact not even enough. Will return with a more elaborate PR.

piitaya:
It's an advanced usage. May be we can add a variable in the mushroom theme and get it in JS. I will give it some try.
The reason we choose 10 is for usability on mobile device to avoid trigger slider when scroll in iOS for example.

Hooolm:
Yeah i agree that the threshold makes sense. I just overlooked the tap in hammer.js.

Consider adding this in.

src/shared/slider.ts
```diff
@@ -104,6 +107,18 @@ export class SliderItem extends LitElement {
...
+                const percentage = getPercentageFromEvent(e);
+                this.value = this.percentageToValue(percentage);
+                this.dispatchEvent(
+                    new CustomEvent("current-change", {
+                        detail: {
+                            value: Math.round(this.value),
+                        },
+                    })
+                );
+            });
```
piitaya, line 121: I think a dispatch change event is missing here (like in the move end event)
```
this.dispatchEvent(
    new CustomEvent("change", {
        detail: {
            value: Math.round(this.value),
        },
    })
);
```

piitaya (changes requested):

piitaya:
I just pushed another PR to add threshold customization if needed #219

src/shared/slider.ts
Hooolm: I agree and that should do it. Did not see that the difference in change and current-change.

Hooolm:
I will look into the threshold config

Hooolm:
Not sure how to test with the new config, other than the default value works, and tap works!

So in my eyes they work more snappy now without changing existing "swipe-gaurds".

piitaya:
Thanks for the changes. No need to merge my branch into yours. I will merge the separate two PR 🙂

PR #221, opened 2022-04-01
add Italian Translation

alessandroias:
Adding an Italian translation.
 - Created the italian translation file into `it.json`. 
 - Reordered languages in alphabetic order into `localize.ts`.


piitaya:
Looks good to me 🙂 Thanks!

PR #224, opened 2022-04-02
Add  dutch translation

Metal-Eagle:
Added dutch translation

### DIFF
--- a/src/localize.ts
+++ b/src/localize.ts
@@ -21,7 +21,7 @@ const languages: Record<string, unknown> = {
     "pt-BR": pt_BR,
     sv,
     "zh-Hans": zh_Hans,
-    "nl": nl,
+    nl,
 };
 
 const DEFAULT_LANG = "en";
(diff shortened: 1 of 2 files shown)


src/localize.ts
```diff
@@ -19,6 +21,7 @@ const languages: Record<string, unknown> = {
     "pt-BR": pt_BR,
     sv,
     "zh-Hans": zh_Hans,
+    "nl": nl,
```
piitaya, line 24:
```suggestion
    nl,
```
No need to repeat the `nl` key.

piitaya (changes requested):
Can you fix the conflict with the main branch ? It's good for me, I will merge when it's done 🙂

thibmaek:
As a native Dutch speaker lots of these seem incorrect to me. 
For example:
- laatste gewijzigd should be laatst gewijzigd
- horizontaal layout should be horizontale layout
- justify is uitlijnen in dutch

src/localize.ts
Metal-Eagle: I edited this, it should look better now.

piitaya (approved):

PR #232, opened 2022-04-04
feat(slider): disabled actual control when unavailable rather than only styling to appear disabled

ulic75:


src/cards/light-card/controls/light-color-temp-control.ts
```diff
@@ -25,10 +27,16 @@ export class LightColorTempControl extends LitElement {
...
+        if (state === "off") {
+            sliderStyle["--gradient"] = "none";
+        }
+
         return html`
             <mushroom-slider
+                style=${styleMap(sliderStyle)}
                 .value=${colorTemp}
-                .disabled=${state !== "on"}
+                .disabled=${!isAvailable(this.entity)}
```
piitaya, line 39: What about adding a new property `inactive` and replacing the `disabled` class by `inactive` class in the slider ? 
The will be 2 properties for 2 different purpose :
- `disabled` : disable the interaction with the slider
- `inactive` : change style for inactive style for the slider

What do you think about that?


It can avoid to pass sliderStyle to the slider.

```suggestion
                .disabled=${!isAvailable(this.entity)}
                .inactive=${!isActive(this.entity)}
```

piitaya (commented):
Nice ! Good idea 🙂 
I just suggest some improvement. Tell me if it seems coherent for you 🙂

src/cards/light-card/controls/light-color-temp-control.ts
ulic75: Seems good to me.  If can't carve out some time today I should be able to make those changes tonight.

piitaya: Yep we can rename it to `--bg-color-inactive` and `--main-color-inactive` for consistency 🙂

ulic75:
The discussed changes have been implemented in the commits above.

piitaya (approved):
Great 🎉

PR #239, opened 2022-04-06
Support getting the isActive state of plant domain entities

jacek-dargiel:
This allows to signal problems with plants on the chips cards using color.

plant chips


piitaya (approved):
Didn't known there is a plant domain 😅 Thank's !

Issue #107, opened 2022-02-13
A climate card


hcoohb:
Is there any plan for a climate card?
Love the mushroom cards and a climate/thermostat card would be a great addition to the current cards.
As the official HA is really huge and very different looking on the dashboard.

Thanks!

brandensilva:
The README docs indicate there is a climate card planned in the future. 

I think we should move towards this look which is where the mushroom design is inspired from with Yuhang's design work on Behance:

<img width="748" alt="image" src="https://user-images.githubusercontent.com/200190/154092599-d7a9f92d-a3e7-4475-ba27-1fdd17a5d256.png">

The + and - are a nice addition over the typical climate change cards because I find it to be rare to extend the temperature more than a couple degrees for my use cases. This is way easier to tap a couple times than trying to move the circular inputs.

hcoohb:
Oh, sorry, I missed the line on the readme. My bad.

Now if we are discussing design, like you said, use case to modify temperatures by ore than a few degrees are probably extremely rare.
So personally I do not think the design from Yuhang that keeps the sliding circle around the value makes much sense and it visually breaks from other controls.
To keep in line with current design of light card, I think, we could have the controls on a second line.
Selecting a prefered mode on the card ui editor, let you turn on/off by clicking on the icon. And then if you want to switch to other mode, icons on second line let you do so.

Here is what I coded using button-card for now, that matches my other mushroom-card themes:
![image](https://user-images.githubusercontent.com/12975783/154389218-3fe0ab75-a612-4a3d-80c8-00b19e9ae39a.png)
![image](https://user-images.githubusercontent.com/12975783/154388381-827fc2a2-43c9-49bb-b809-c04fdbcf716e.png)

zsarnett:
I would recommend not working with the Round Slider. (as the one who built it for HA its not worth it and the UX isn't great)

brandensilva:
@zsarnett Yeah it's a nice visual but I can definitely see how the UX isn't strong. What I do like about it perhaps though is it gives a wee bit of perspective from a top view of the low and highs and where the current temperature target falls. It doesn't help you any really in knowing what to set the temp, but it gives perspective. 

@hcoohb This seems like a good start. I think part of the goal should be to get something unique looking still if we can so it stands out from all the other cards while maintaining usability. 

[...]

PR #258, opened 2022-04-08
feat(thermostat): initial climate support

ulic75:
I apologize if I'm out of line.  I guess I'm impatient.  I tried to follow coding styles but I'm sure I've missed things.  I'm willing to address any suggestions as quickly as I can.  I decided to call the card thermostat only because that's what the official HA card is called, I'm not opposed to renaming it to climate (I actually started with it as climate and then switched to thermostat).  I can also pretty easily split the slider changes and the thermostat card into two PRs if you'd prefer (obviously one would be dependent on the other).

https://user-images.githubusercontent.com/5641964/162476130-18e9dae6-fae4-4754-aa88-ce95ebe43d61.mov

### Slider Changes
- Add support for secondary value which fills from right to left
- Add support for a gap property which limits how close value and secondary can be to each other
- Add support for step property which dictates increments when sliding

### Thermostat Card Features
- Supports all features that the official HA thermostat card supports plus some
- Displays current state and preset modes
- Separate target temperature controls for high and low
- Changing of modes (heat, cool, off, etc.)
- Display of icon/color for the current running mode (heating, cooling, drying, idle) or state if running mode not supported
- Optional Energy Efficient Gap, prevents high/low collision and mode cycling, so heat until it cools which then turns on the heat

### Todos
- [x] thermostat card
- [x] thermostat card editor
- [x] thermostat card color rules
- [x] documentation
- [x] hass_dev

Related #107

P.S. The commits are a mess, sorry if you try to follow any kind of flow in them 😄


ulic75:
Sure of course.  It's your project 😄   I just wanted something myself and not sharing it didn't seem right.  I'm happy to help further or not, whatever seems best to you.

KrX3D:
Hello, will this card be able to add more sensors to show? something like i use for now:

image


thanks so far, the card looks nice

piitaya:
No, not for now. Current card will be simple to answer to most users needs (just hvac + temperature control).
We will look how we can provide additional information but not in short term because the current mushroom card are not big so we can't display lot of information.

ulic75:
@piitaya Was there anything in particular you wanted me to do with this PR?

CallMeAreks:
Is this card still in development? Looks great and pretty much exactly what I need. If there's still things left to develop/review. how can I help?

ulic75:
> Is this card still in development? Looks great and pretty much exactly what I need. If there's still things left to develop/review. how can I help? @ulic75 @piitaya

I'm not sure where things are at.  You can install a "single" version of this card at https://github.com/shroom-singles/thermostat until there is something more official.

Schluggi:
Nice work :)

I've just one question: Why you renamed _climate_ to _thermostat_ in f071eb1? Wouldn't it be nicer to stay with climate as name to be analog to the build-in climate card?

ulic75:
I chose thermostat because that is the name of the official HA card that performs the same function. It could be renamed to climate.

zsarnett:
So I have been using this card for a week or two now, and I think that the card looks amazing. but.... it suffers from the same issues the native Thermostat Card has. 

The main one is that when using a thermostat, in my case (and I would assume most), I use it to change the temperature 1-4 degrees. This means that I am going from 74 degrees to 72 degrees. And that changes the feel of the room tremendously.

Sliders have issues when it comes to this. Moving slight increments is not easy. Most of the time you find your self jumping between 71-73 degrees instead of hitting the 72 because the distance is so small between ticks.

I have that issue and also that if you start holding the slider it actually starts you out at the degree you taped which will sometimes change the number immediately with ever moving.

Again, I think that the card looks great and the code quality is very good. But the UX of this just falls short the same way the thermostat card I built for core HA did...

I think the design has to have buttons to increment the temperature else we will run into the same issue.

How to do this? I am not sure yet but I am sure we can figure it out

piitaya:
Thanks for the feedback @zsarnett 😍 I have underfloor heating so I did not use temperature change 😅
The thermostat card is highly awaited so I prefer to be sure before release.
For me, the main issue is that the current design can't fit in a grid. I wonder if the thermostat card should only offer hvac mode and preset changes keep the more-info dialog to adjust the temperature. Or just propose plus and minus button to adjust the temperature instead of the slider?
And I don't known how to manage the `heat-cool` mode 🙁 Managing 2 temperatures from a small card can be difficult.

zsarnett:
Yea I think that it's definitely an option to move the temp change to the more information dialog

I also like the idea of 3 buttons on the bottom.

Temperature, Up, Down

This will work for anything other than `heat-cool`

Heat cool could add another button that would switch to change the temp of heat or cool? 

but I agree. Climate is a hard one that need to get right before releasing

CallMeAreks:
How about something like the thermostar card in UI Lovelace Minimalist? (image)

ulic75:
I've got some ideas mulling around in my head, just no time to work on them 😢

alexreddy78:
Dear @ulic75 @piitaya 

using the shrooms thermostat card, find flw bugs: main icon of entity is always grey, whether is active and thermostat is on, or inactive and thermostat is off. And second one: card background is transparent after update to 1.7.3.

Kindly check and advise

realroywalker:
> Thanks for the feedback @zsarnett 😍 I have underfloor heating so I did not use temperature change 😅 The thermostat card is highly awaited so I prefer to be sure before release. For me, the main issue is that the current design can't fit in a grid. I wonder if the thermostat card should only offer hvac mode and preset changes keep the more-info dialog to adjust the temperature. Or just propose plus and minus button to adjust the temperature instead of the slider? And I don't known how to manage the `heat-cool` mode 🙁 Managing 2 temperatures from a small card can be difficult.

It's good that you are focussing on the right solution before release, I find the default HA card not in-keeping with the rest, it's so huge!
I like the sound of plus/minus to keep it small, maybe the HVAC mode could temporarily "replace" those buttons when a mode button is pressed, to keep the whole thing small?
To be clear I'm talking about a "mode" button on the card that toggles between setpoint buttons and the actual HVAC/climate mode.

gogui63:
Hello, any news about this PR? I am very interested in this feature to finish my dasboard :)

piitaya:
> Hello, any news about this PR? I am very interested in this feature to finish my dasboard :)

We try to find the best way to handle the climate action. We are not convinced about the slider usage.

ulic75:
@piitaya Feel free to close and/or lock this PR if you want.

alexreddy78:
@ulic75 @piitaya 
Idea is good! Card is also fine (shroom’s one) I am using it and have no bugs or problem. Why PR should be closed or locked, Dear DEVELOPERS? It better to merge and allow to use it for everybody. It’ better than nothing :)))

Regards

liquidbear99:
Any idea if this will be merged soon or is this card a long ways off? I'm happy to go with the option above [https://github.com/shroom-singles/thermostat ](url) if needed!

mihailescu2m:
After playing with my own thermostat custom card, I have to agree about the slider, it is not ideal.

Screen Shot 2022-06-26 at 8 57 16 am


My 2c after many iterations (different from the above ^^^):
primary: [name] - [current temp] - e.g. Living - 19°C
secondary info: [hvac_action] to [target temp] - e.g. Heating to 20°C
icon color: conditional on hvac_action: orange (heating), blue (cooling), green (auto), yellow (dry), gray (fan/idle/off)
collapsible row with target controls: (decrease button) (target temp) (increase button) (mode button) - e.g. - 20 + [mdi:fire]
mode button is dynamic and shows the current hvac_action (mdi:fire, mdi:snowflake, etc...)
pressing mode button changes row with target controls in row with mode selection buttons (like the video in the first comment).

PR #261, opened 2022-04-08
fix(shape-icon): use box-shadow instead of outline

### LINKED ISSUE #245: Icon border / outline changes to square when color temp is below 180
**Describe the bug**

When adjusting the color temperature of a light, if the color temp is set to below 180 the icon changes to square rather than circle
This issue seems to affect Safari but not chromium based browsers

**To Reproduce**
Adjust a light to below 180 color temp.

**Screenshots**

Safari:
Screenshot 2022-04-07 at 10 33 42

Chromium (Vivaldi):
Screenshot 2022-04-07 at 10 38 42

ulic75:
Safari apparently doesn't support round outlines with border-radius.  This achieves the desired effect while still being round.

Closes #245

### DIFF
--- a/src/shared/shape-icon.ts
+++ b/src/shared/shape-icon.ts
@@ -44,11 +44,11 @@ export class ShapeIcon extends LitElement {
                 align-items: center;
                 justify-content: center;
                 background-color: var(--shape-color);
-                transition-property: background-color, outline;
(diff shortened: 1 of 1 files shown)


piitaya (approved):
Great! 😜 I don't why it was outline 🤷‍♂️

PR #263, opened 2022-04-09
feat(light-card) selection limit for light brightness

skrynklarn:
Other sliders cards have ability so set a limit, min and max value, that user can select in the slider.

Example of use case:
Set min to 1, so it will be easier to the user to set 1% instead of 0 and turn off the light by accident.


skrynklarn:
Update my branch and fixed merge conflicts.

@piitaya Will you have a look on this feature when you have time? Thanks

piitaya:
Do you have another use case for the min and max ? I'm not sure it's a good idea to add 2 new options for this. 
Mushroom must stay simple. I prefer an option to avoid turn_off the light instead of setting a min / max.

IceBotYT:
> Do you have another use case for the min and max ?

Some lights (e.g. smart dimmers) don't change brightness after a certain value. `85% == 100%` for example.

skrynklarn:
I use it for almost all my lights where I can change brightness so as before set min: 1 so it is not possible to turn it off with the slider. I also use it some some light to limit max brightness for me and also on some dashboards for my kids so they can not play as they like :) It could be useful for example if there is a volume slider to limit max (I know this PR is only for light)

No more use cases for me. I understand that mushroom is to be simple and I trying to use it instead of UI minimalist. But some small features missing and this was one for me. 

One option would be to have that settings as ymal only. But that breaks the idea of mushroom as a ui config cards. But maybe for advanced/special settings that could be a option?

Don't know if it is possible but a collapsable "advanced settings" would be nice in card settings but have not seen any cards have it. Maybe something for Home assistant to add or if it can be done for custom cards.

Just some thoughts.

Mrjwojc:
Hi @skrynklarn Is your code merged? And if so, how to I use it when my Mushroom is at the current version?

dannybloe:
I think this pull request solves one of the biggest problem I have with most sliders controlling lights and that is that it is almost impossible to set levels to very low percentages. I never ever want the option to set it to 0% just to turn off the lights. I never do that. Switching off the lights is always a toggle operation for me and never a slider operation. I have many people in my family complaing that they cannot use the slider to set low levels. Having that said, for this usability problem it would be good enough for me just to **remove** the option to set the level to 0% all together (min = 1) and have no setting for that. That would still keeps things simple. In fact: even more simpler because now you can use the lower range of the slider.

RickeyWard:
I agree min/max are nice options for all sliders. My dimmer switches let me hardware set the max and min but I still would prefer to set most of the lights to 1% easily. I currently do that with call_services on doubletap action but friends and family don't know all my secret shortcuts.

Snellingen:
The feature would be very welcome and is something that will fix #847 and feature #1217 (partially). 
I have some lights that needs a max, but the most critical fix I need is to avoid turning off the lights when sliding to min brightness. This PR or either #848 will fix this. What is currently blocking these and what can we do to move this along?

hanseandersson:
Im also looking for this!
Especially for light card to be able to slide all the way down and stop at 1% (either det as an option or atleast be able to yaml code it)

Also for media card to be able to set max volume to like 30% so kids won't get deaf when they control their own music 😂😅

Matsched:
Can we do something to push this topic a little bit?
Unfortunaly i have no clue of programming, but maybe the MR from [skrynklarn](https://github.com/skrynklarn) can do it? I would donate for getting the feature that my sound system cant go over 50% (https://github.com/piitaya/lovelace-mushroom/issues/1217)

e965:
Hey @piitaya This seems to be a sought-after change (I'd like to implement it too)

PR will soon celebrate its third anniversary lol
Maybe you can do something to push it forward?

piitaya:
I'm closing this one in favor of https://github.com/piitaya/lovelace-mushroom/pull/848/
Mushroom should stay simpl with enough options but not too many so I do not prefer to add custom limits for that.

e965:
@piitaya you shut down a request with 22 likes and a bunch of likes under almost every comment here, in favor of a solution with 8 likes (and which doesn't solve the majority's problem)

Full understanding 🙂

Jonybat:
> @piitaya you shut down a request with 22 likes and a bunch of likes under almost every comment here, in favor of a solution with 8 likes (and which doesn't solve the majority's problem)
> 
> Full understanding 🙂

I don't mean to start an argument here, but I don't think it is acceptable to demand or question decisions about something that we are getting for free. Their project, their choices. 

You see this a lot in the open source world, and even more so in projects connected with tools that start touching the mainstream (like homeassistant), but the only thing it achieves, is pushing the great people behind these projects away.

e965:
@Jonybat I have no problem with whatever decision was made by the project maintainer, I'm just sad that it's not the one that the objective majority would have wanted

If the maintainer thinks it's better, then okay.

piitaya:
The only example of use-case than can be implemented right now with the PR is : 

> Set min to 1, so it will be easier to the user to set 1% instead of 0 and turn off the light by accident.

This use case is solved by https://github.com/piitaya/lovelace-mushroom/pull/848 without user configuration. I have no feedback for user than want to limit the brightness to X%.
I agree that they could be a use-case for limiting the volume but it's not the purpose of this PR and only the max should be configurable. Mushroom's success is due to the fact that it is easy to configure and you don't have tons of options for 0.1% of users.

In the near future, Mushroom will support [card features](https://www.home-assistant.io/dashboards/features/) so you will be able to plug custom features like [Custom Features for Home Assistant Cards](https://github.com/Nerwyn/service-call-tile-feature) so you can build slider with the rule you want.

e965:
@piitaya Thanks for the detailed explanation of your motivation, it's much clearer now

I use Mushroom myself, and I hope that the project will only get better in the future.

PR #307, opened 2022-04-14
fix: do not consider unavailable if entity state in unknown

### LINKED ISSUE #298: Cover card disabled if unknown status
Hello again.
I use Somfy RTS motors for my covers and they do not report the state to home assistant. 
After each restart of home assistant they go into state unknown until first use.  
Every time they go into an unknown state mushroom cover card commands are disabled.

To reactivate the card I have to use the home assistant commands once to change the state of cover.

Can be fixed this behavior please?
Thank You

piitaya:
fix #298


srodighiero:
Hi, is there any way to rename the status via YAML? I'm thinking about covers, where it would probably look better something different (or nothing) rather than "unknown" status

Issue #331, opened 2022-04-15
Alarm control panel not working anymore


tnagels:
I use the Satel Integra integration for control of my alarm. Since the Upgrade of Mushroom to 1.6, I can't switch the alarm on/off anymore. The buttons to control the alarm are greyed out and don't do anything when I click them. 

Switching on and off with the Home Assistant built-in control panel (reached by clicking more info) does work as expected.

larry-glz:
Same here w Envisalink and Arlo.

JohnnyM84:
Same for me

Osjuh:
Same for me, with Somfy.

hebbister:
Same for me

PR #336, opened 2022-04-16
fix(alarm): available

acesyde:
Fix : #331


acesyde:
@piitaya Ready to review

Issue #348, opened 2022-04-16
Person Card - Away icon logic incorrect


greghesp:
**Describe the bug**
The person card seems to have 3 icons. Home, Office and Away.
I haven’t looked at the code, but I’m assuming that if the state isn’t Home or Office then it defaults to showing the away icon.

Update: https://github.com/piitaya/lovelace-mushroom/blob/af3f20c051eb5df6607a8ce09d7df94728d41ab8/src/cards/person-card/utils.ts#L15
I would suggest here to check to see if state is not_home, then return away icon. Otherwise, check return home icon. Then check zone icons

I am using a device tracker which integrates with room presence. Therefore the state of my tracker can be things such as Living Room.  If this is the state, then the card icon shows the person as away when they’re not

**To Reproduce**
Set person card entity to anything apart from home or office

**Expected behavior**
Icon should only show away icon is entity state is not_home

PR #353, opened 2022-04-17
Amend person card logic to allow for room presence

greghesp:
Fixes https://github.com/piitaya/lovelace-mushroom/issues/348

I've adjusted the logic so that if the presence is "not_home", that is the only way the "not home"/`home-export-outline` badge icon can be shown.

It should then continue and check for zone icons, and will then default to the home icon. 
Should users like myself use custom device_trackers which combine home/away presence as well as room presence within the home, they will get the correct device badge icon

### DIFF
--- a/.hass_dev/views/person-view.yaml
+++ b/.hass_dev/views/person-view.yaml
@@ -66,6 +66,15 @@ cards:
             dev_id: demo_anne_therese
             location_name: unknown
       - type: button
+        name: Living Room
+        icon: mdi:sofa
+        tap_action:
+          action: call-service
+          service: device_tracker.see
+          service_data:
+            dev_id: demo_anne_therese
+            location_name: Living Room
+      - type: button
         name: Office
         icon: mdi:office-building
         tap_action:
--- a/src/cards/person-card/utils.ts
+++ b/src/cards/person-card/utils.ts
@@ -5,14 +5,18 @@ export function getStateIcon(entity: HassEntity, zones: HassEntity[]) {
     const state = entity.state;
     if (state === UNKNOWN) {
         return "mdi:help";
+    } else if(state === "not_home") {
+        return "mdi:home-export-outline"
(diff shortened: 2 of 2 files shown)


piitaya:
How do you track using custom device_trackers ? Can you provide example of state of your person entity?

greghesp:
> How do you track using custom device_trackers ? Can you provide example of state of your person entity?

I need to double check my logic on this again, as I found a big in my device tracker.

What I've done is created a custom device_tracker entity that essentially looks at the person entity to see if there home or away, and then the sensor entity from the `mqtt_room` integration.

If their person entity is home, then it will push the room location to the `device_tracker.see` command, otherwise it'll push the location from the person entity

piitaya:
I don't understand. What is the state of your entity ? Device tracker and person entities can take `home`, `not_home` and zone names as state. 
If you want the icon of your zone, just set an icon from Home Assistant settings.

greghesp:
device_tracker can accept any state, like zones. Not just home or not_home.

The issue is that if you're using a beacon for room presence,. The current code sets the icon as the away icon rather the home icon. I've changed that so it defaults to the home icon and colour, but does the home and zone check first

https://www.home-assistant.io/integrations/device_tracker/#device-states

piitaya (approved):

piitaya:
fixes https://github.com/piitaya/lovelace-mushroom/issues/348

briandonahue:
I'm using the icloud3 device tracker and it's put people in the "[Stationary](https://gcobb321.github.io/icloud3/#/chapters/2.3-special-zones)" zone when you are outside the house but not moving for a period.  In the person card this shows the home icon which is confusing. Would be nice if it could recognize statrionary as "away". I realize this may be unique to this tracker, so not sure what the best way to handle it would be...

TravisWilder:
> 

I have the same issue - it is not very nice this way

Issue #351, opened 2022-04-17
Color for lock UNLOCKED.


danch99:
Hi,

I would like to make the lock state UNLOCKED more obvious in the UI. Right now when the lock is LOCKED we can define a color, but when it is UNLOCKED the icon is just grey.

In order to clearly see when a lock is UNLOCKED I would like to be able to define a color for the UNLOCKED state.

By default I imagine green when it is LOCKED and red when it's UNLOCKED.

Thanks a lot for your help and the great work !
Mushroom is really a game changer !

piitaya:
There already is an issue about lock state color https://github.com/piitaya/lovelace-mushroom/issues/335. I agree that red/green can be better 😉

danch99:
Great, I haven't seen it. Thanks

FrankJaco:
Hi All,
I see in the current release we do now have color added to Locks.  I think that is great.  But I am in disagreement with danch99 as to what color should represent locked vs unlocked.   To me RED means stop, thus a locked door means you cannot go forward.  Green means go.  You can go through an unlocked door.  I would prefer the colors swapped.  I would like to go further to suggest this paradyne be carried forward to other cards.  For example an open cover (like a Garage door) should show as Green as you can go through them, and Red when the door is closed as you are are stopped from going through the door as it is closed.  An alarm card shows Red when it is alarmed (as you cannot go through the house), and Green with disarmed as you can freely go through the house. Having a consistent color interface would augment useability in my opinion.

If others disagree with my stop/go Red/Green concept, perhaps the ability to set whatever color you want for each state would be a secondary option.

In any case thanks to you and your growing number of helpers for making the Mushroom suite of cards.  They are absolutely the best thing to come to Lovelace.  A final question if you don't mind....  How did you come up with the name "Mushroom" ? 

Thanks for all,

fj

piitaya:
Hi 🙂
Thanks for your feedback. I don't have a lock so I don't known what is the best solution about colors 😅
Your can customize the lock colors in your `theme.yaml` file
```yaml
--mush-rgb-red: 244, 67, 54;
--mush-rgb-pink: 233, 30, 99;
--mush-rgb-purple: 156, 39, 176;
--mush-rgb-deep-purple: 103, 58, 183;
--mush-rgb-indigo: 63, 81, 181;
--mush-rgb-blue: 33, 150, 243;
--mush-rgb-light-blue: 3, 169, 244;
[...]

PR #363, opened 2022-04-19
feat(lock): improve color rules for lock card

acesyde:
Feature : #351

- [x] Doc updated


acesyde:
@piitaya ready to review

piitaya:
Not sure we want to propose this. I think it's better to do this with a theme.

acesyde:
hmmm I understand what you want to do ;)

acesyde:
@piitaya like that ?

piitaya (approved):

Issue #371, opened 2022-04-20
Missing Connection Import


alessandroias:
https://github.com/piitaya/lovelace-mushroom/blob/28ca656eb1d478c826d1822ed9e6514d6c608662/src/cards/title-card/title-card.ts#L2

You should also import `Connection`!
This Line should be:
```ts
import { Connection, UnsubscribeFunc } from "home-assistant-js-websocket";
```

Issue #369, opened 2022-04-20
Cover card icons do not follow HA device class


gadgetbazza:
**Describe the bug**
The cover card always shows the two movement actions as Up or Down arrows.  However, depending on the device class as defined in the "Show As" field of a cover entity, it could be that the device is a curtain or gate.  When a device is configured as a curtain for example it changes the up / down actions on the entity to open / close actions and uses different icons accordingly to the device class definitions.

**To Reproduce**
1. Configure a cover device as a device type of curtain rather than blind.
2. Add a cover card for the curtain entity.
3. Include the control buttons on the card. 
4. You will see the controls show as up and down arrows.

**Expected behavior**
As above, except for step 4 the controls should be open / close arrows as per the HA implementation of device classes.

**Screenshots**
Screenshot showing temporary dashboard with a cover entity configured as a blind and a second configured as a curtain.  Screenshot has Mushroom Cover cards on the left and HA Entities card on the right.

<img width="1032" alt="Screenshot 2022-04-20 at 22 20 12" src="https://user-images.githubusercontent.com/99490699/164324984-869e0932-d0f5-491c-8483-f04b50ad38f3.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Safari
 - Version 15.4

PR #372, opened 2022-04-20
Tiny Bugfix + Updated Italian translation

alessandroias:
- Fix to issue #371 :
    Importing `Connection` in `title-card.ts`.
- Fix to issue #369  :
Displaying `mdi:arrow-collapse-horizontal` and `mdi:arrow-expand-horizontal` when the cover's device class is `curtain` as per HA's entity card.

### DIFF
--- a/src/utils/icons/cover-icon.ts
+++ b/src/utils/icons/cover-icon.ts
@@ -1,4 +1,5 @@
 import { HassEntity } from "home-assistant-js-websocket";
+import { state } from "lit/decorators";
 
 export const coverIcon = (state?: string, entity?: HassEntity): string => {
     const open = state !== "closed";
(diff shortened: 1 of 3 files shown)


src/utils/icons/cover-icon.ts
```diff
@@ -1,4 +1,5 @@
 import { HassEntity } from "home-assistant-js-websocket";
+import { state } from "lit/decorators";
```
piitaya, line 2: Unused import 😅

src/utils/icons/cover-icon.ts
```diff
@@ -100,8 +102,10 @@ export const computeOpenIcon = (stateObj: HassEntity): string => {
 };
 
 export const computeCloseIcon = (stateObj: HassEntity): string => {
+    console.log(stateObj.attributes);
```
piitaya, line 105: This should be removed too

piitaya (changes requested):
Thanks ! 🙂

piitaya (approved):

PR #374, opened 2022-04-21
feat(media-player): show volume next to state when not unavailable/unknown/off

### LINKED ISSUE #352: Media player card - Show volume level instead of "on"
First of all thanks for this great project. Thanks to you my dashboard looks much more pleasing now.
I would like to ask if it is possible to change the "on" in the secondary info section of a media player card to display it's volume.

It's nothing game breaking but it makes the interface feel inconsistent.

Currently it looks like this:
image

This is what I would expect/suggest it to look like:
image

**Describe alternatives you've considered**
Alternatively we could add template support for the secondary information.

**Additional context**
In my opinion this change would make the UI feel more consistent since lights and other components handle it like this. Here are some examples:
image
image

Thank you for reading so far and maybe even for considering it ;)

madmurl0c:
Hey there,

this is just a little pull request that would display the current volume level next to the state of a media player.
image

Fixes #352 

Please let me know if there's something missing.


acesyde:
Hi @madmurl0c,

You can add this function https://github.com/piitaya/lovelace-mushroom/blob/47261608232152fa834053ee648b1eae4628166a/src/cards/fan-card/fan-card.ts#L89 and the connection to the slider to update the percentage when the user change the volume

madmurl0c:
Hey @acesyde,

thank you very much for your feedback.
Integrating it the way you mentioned would really be much better so I gave it a try.
I took inspiration from the light-card since it was easier for me to understand but as far as I've tested it there weren't any issues but I'd be glad if you could take a look at it :)

acesyde:
@madmurl0c that looks good

madmurl0c:
Thanks for taking a look at it

apivaral:
That's perfect!

piitaya (approved):
Looks good to me 🙂

Issue #354, opened 2022-04-17
Fan Reverse Button


muddro1:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
in the current fan card you allow for oscillation. Would love to see an option for fan direction too in place of the oscillation.

PR #382, opened 2022-04-22
Add fan direction functionality

greghesp:
## Description

3rd time lucky now i've sorted my Github issue.  This PR adds the fan direction functionality

## Related Issue

https://github.com/piitaya/lovelace-mushroom/issues/354

## Motivation and Context

Adds a core functionality feature for the fan entity

## How Has This Been Tested

No real world testing, but tested in development environment and confirm device attribute changes

## Types of changes

## Checklist

### DIFF
--- a/src/utils/entity-styles.ts
+++ b/src/utils/entity-styles.ts
@@ -20,6 +20,14 @@ const strAnimations = {
             transform: rotate(360deg);
         }
     }`,
+    spinReverse: `@keyframes spinReverse {
+        from {
+            transform: rotate(0deg);
+        }
+        to {
+            transform: rotate(-360deg);
+        }
+    }`,
 };
 
 export const animation = {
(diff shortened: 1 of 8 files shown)


src/utils/entity-styles.ts
```diff
@@ -20,6 +20,14 @@ const strAnimations = {
             transform: rotate(360deg);
         }
     }`,
+    spinReverse: `@keyframes spinReverse {
```
alessandroias, line 23: Wouldn't it be better to use CSS' [`animation-direction`](https://developer.mozilla.org/en-US/docs/Web/CSS/animation-direction)  instead of having two different animations? 
It would be cleaner, more readable and most importantly, less code to maintain.

greghesp: That would make sense.  Updated!

piitaya:
Hi 🙂 
Don't you think it can be better to add a list of actions like the vacuum card ?

NewsGuyTor:
> That just leaves `fan.set_preset_mode`[...]. I'm not sure how `set_preset_mode` would even work here?

For Xiaomi smart fans you can choose between natural and constant fan speed. 

Natural is way more pleasant IMO than the exact same fan power all the time  (like old fashioned fans).

Issue #126, opened 2022-02-15
Optimize Card Light


cyrilphoenix71:
**Is your feature request related to a problem? Please describe.**
I this page we can see option of the Minimalist Light Card : https://ui-lovelace-minimalist.github.io/UI/usage/cards/card_light/

**Describe the solution you'd like**
Integrate the card_light_slider_collapse and card_light_slider_horizontal options would be cool

piitaya:
This issue already exist : https://github.com/piitaya/lovelace-mushroom/issues/34. I will close this one.

cyrilphoenix71:
I know but minimalist is not user friendly and this only two option are interresting I thing

PR #383, opened 2022-04-22
feat(light): add collapsible controls for light

M4D1NG3R:
## Description
implementation of the "Collapse slider when off" function from [Ui Lovelace Minimalist](https://ui-lovelace-minimalist.github.io/UI/).

## Related Issue
[#126](https://github.com/piitaya/lovelace-mushroom/issues/126)

## Motivation and Context
I really liked the function & implemented a workaround with the state-switch card, but thought it would be great to have this function natively.

## How Has This Been Tested
I tested it with the Development server as mentioned in the README

## Types of changes

## Checklist

I could need some help with translations, testing & documentation.
Also, I tried to align with the code style of this project, but I am open for critics.

### DIFF
--- a/src/cards/light-card/light-card.ts
+++ b/src/cards/light-card/light-card.ts
@@ -141,14 +141,16 @@ export class LightCard extends LitElement implements LovelaceCard {
         if (!entity) return;
 
         const controls: LightCardControl[] = [];
-        if (this._config.show_brightness_control && supportsBrightnessControl(entity)) {
-            controls.push("brightness_control");
-        }
-        if (this._config.show_color_temp_control && supportsColorTempControl(entity)) {
-            controls.push("color_temp_control");
-        }
-        if (this._config.show_color_control && supportsColorControl(entity)) {
-            controls.push("color_control");
+        if (!this._config.enable_collapse || entity.state == 'on'){
+            if (this._config.show_brightness_control && supportsBrightnessControl(entity)) {
+                controls.push("brightness_control");
+            }
+            if (this._config.show_color_temp_control && supportsColorTempControl(entity)) {
+                controls.push("color_temp_control");
+            }
+            if (this._config.show_color_control && supportsColorControl(entity)) {
+                controls.push("color_control");
+            }
         }
         this._controls = controls;
         const isActiveControlSupported = this._activeControl
--- a/src/cards/light-card/light-card.ts
+++ b/src/cards/light-card/light-card.ts
@@ -199,61 +199,47 @@ export class LightCard extends LitElement implements LovelaceCard {
 
         const rtl = computeRTL(this.hass);
 
-        const content =  html`
-            <mushroom-state-item
-                ?rtl=${rtl}
-                .layout=${layout}
-                @action=${this._handleAction}
-                .actionHandler=${actionHandler({
-                    hasHold: hasAction(this._config.hold_action),
-                    hasDoubleClick: hasAction(this._config.double_tap_action),
-                })}
-            >
-                <mushroom-shape-icon
-                    slot="icon"
-                    .disabled=${!active}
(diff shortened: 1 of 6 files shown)


piitaya:
Hello! 
I tried to implement it in the past but there were issue will the grid layout :


The card take the full height so the option is not very useful and the UI is not consistent. I think we need to fix this issue before adding the issue.
Also, it can be an option for all cards, not just for lights. I will look how we can refactor the code to reduce the copy paste between card 🙂

Thanks I will prioritize this feature 🙂

greghesp:
I understand why you want to do this, but it definitely needs to be an option.  Having the slider visible whilst the light is off, means you can tap the slider at any position which turns it on at that brightness.  Without that, you can only turn it on/off

M4D1NG3R:
@piitaya thanks for the feedback, I only tested it without the grid:
image
image

Moreover, I only partially understand your problem. If I have two lights, one with and one without a slider, the card will also take the full height. So the problem exists independently of my implementation:
 image

In addition, when both lights are off, the size is reduced. So the goal of saving space is fulfilled:
image

@greghesp I implemented it as an option:
image

M4D1NG3R:
@piitaya 
I would really like to contribute, but I am a little confused. Do I understand this correctly:
When measured by your standards, the UI in its current state is already not consistent?
Could you explain, how your preferred version would look like?

M4D1NG3R:
@piitaya please check again I implemented a version which works like this:
image

Does this align with your standards?

M4D1NG3R:
@piitaya did you have some time to look at my implementation?
If so, I would be very happy to get some feedback from your side.

Thank you in advance!

src/cards/light-card/light-card.ts
```diff
@@ -197,47 +199,61 @@ export class LightCard extends LitElement implements LovelaceCard {
...
+                    </mushroom-card>
+                </ha-card>
+            `;
+        }
+        else{
+            return html`
+                <mushroom-card .layout=${layout} ?rtl=${rtl}>
+                ${content}</mushroom-card>
+            `;
+        }
```
piitaya, line 256: This part looks a little bit too complicated. I think it's better to let the layout like it was before.

src/translations/en.json
```diff
@@ -50,6 +50,7 @@
                 "use_light_color": "Use light color",
                 "show_color_temp_control": "Temperature color control?",
                 "show_color_control": "Color control?",
+                "enable_collapse": "Collapse slider when off",
```
piitaya, line 53:
```suggestion
                "enable_collapse": "Collapse controls when off",
```

piitaya (commented):
Hi 🙂 Sorry, the PR was in draft, I thought it was not ready to review 😅

Maybe a `collapse_controls` instead of `enable_collapse` can be more explicit ? What do you think?

I will investigate how we can change the layout to support filling parent height or not. I will give some tests 😅  It will be another PR.

M4D1NG3R:
Hey, many thanks for the feedback!
I reverted the changes regarding the "card takes full height" issue and edited the README regarding your request.

Also, the card is PR is now ready for review.

piitaya:
What about renaming `enable_collapse` to `collapse_controls` ? May be you think `enable_collapse` is better ? or `collapse_controls_when_off`? 😅

M4D1NG3R:
I am open for anything which fits better in your opinion.
I will change it to `collapse_controls`

greghesp:
> A small note: The question came up in the forum why it doesn't work. I am not familiar with the development process here on Github. But it is confusing for users when the readme has already been updated before the corresponding release. That aside: Great addition!

I think this is a HACs thing. Not overly familiar with how HACs reads things, but I believe an Info.md is needed for HACs if you want the Readme to belong only to the repo

Issue #349, opened 2022-04-17
Add a plant card


chrisb86:
I would love to see a plant card. At the moment I'm using "Lovelace flower card" to display the states of my Mi Flora sensors. It displays a picture of the plant and 4 bars for temperature, light, humidity and conductivity. 

The design doesn't fit well in a mushroom dashboard. I would like to have a card with bars like this, but without a picture. 

There exists a minimalist card that modifies the "Lovelace flower card" and makes it nice and compact. This could be an inspiration. 

I tried to build something myself with a template card. But without bars or even the possibility to add icons for the different sensors it doesn't work well. Also the possibility to use 4 cards per plant is not very convenient. 

Lovelace Flower Card:
![image](https://user-images.githubusercontent.com/560571/163713032-00a98e44-86aa-4782-ae2a-62256d104f75.png)

Minimalist Flower Card:
![image](https://user-images.githubusercontent.com/560571/163713144-d3472136-f614-4ac7-b9f5-7dbe163b8e8d.png)

PR #388, opened 2022-04-24
feat(plant): Plant card

greghesp:
## Description

Started work on building out a Plant card which will work with the default [plant](https://www.home-assistant.io/integrations/plant) integration, but also the [custom plant patch](https://github.com/Olen/homeassistant-plant) to pull plant data from Plantbook (mainly species and image)

## Related Issue

Fixes https://github.com/piitaya/lovelace-mushroom/issues/349

## Motivation and Context

The plant entity is not yet supported by Mushroom

## How Has This Been Tested

Still a WIP.  Please review and comment if you see anything that could be improved

## Types of changes

## Checklist

- [ ] I have tested the change locally.


src/cards/plant-card/plant-card-editor.ts
```diff
@@ -35,6 +45,20 @@ const computeSchema = memoizeOne((icon?: string): HaFormSchema[] => [
             { name: "use_plantbook_picture", selector: { boolean: {} } },
         ],
     },
+    {
+        type: "grid",
+        name: "",
+        schema: [
+            { name: "min_temperature", selector: { number: { mode: "box" } }, default: 3 },
```
greghesp, line 52: Does anyone know why the default option isn't pre-populating the fields?

piitaya:
Hi 🙂 Thanks for the contribution. I saw that you are adding lot of option (min_temperature, etc...). There parameters are set in the [plant integration](https://www.home-assistant.io/integrations/plant/#configuration-variables) so we don't need to add them in the card config

greghesp:
> Hi 🙂 Thanks for the contribution. I saw that you are adding lot of option (min_temperature, etc...). There parameters are set in the [plant integration](https://www.home-assistant.io/integrations/plant/#configuration-variables) so we don't need to add them in the card config

I realised that just after my last push.  Any suggestions on what to do should the min/max values not be within the plant entity?

> Also, I think it's better to add `use_entity_picture` option instead of `use_plantbook_picture`. I submitted an issue to use the `entity_picture` field instead of `image` in the custom integration [Olen/homeassistant-plant#17](https://github.com/Olen/homeassistant-plant/issues/17).

This makes sense. I will revert it to `use_entity_picture` for the option, but will keep the logic grabbing the image value for now as the plant integration doesn't seem to support `entity_picture`

piitaya:
Use this code to support both.
```ts
const picture = this._config.use_entity_picture
            ? entity.attributes.entity_picture || entity.attributes.image
            : undefined;
```

> Any suggestions on what to do should the min/max values not be within the plant entity?

Why would these values not be present? If they are not present what is the purpose of plant monitor integration? May be we can hide indicator without min/max?

greghesp:
> Use this code to support both.
> 
> ```ts
> const picture = this._config.use_entity_picture
>             ? entity.attributes.entity_picture || entity.attributes.image
>             : undefined;
> ```
> 
> > Any suggestions on what to do should the min/max values not be within the plant entity?
> 
> Why would these values not be present? If they are not present what is the purpose of plant monitor integration? May be we can hide indicator without min/max?

I've noticed they don't seem to show for some reason.  For example:

```yml
plant:
  Tulip:
    sensors:
      moisture: sensor.plant_moisture
      battery: sensor.plant_battery
      temperature: sensor.plant_temperature
      conductivity: sensor.plant_conductivity
      brightness: sensor.plant_brightness
    min_moisture: 15
    max_moisture: 60
    min_temperature: 6
    max_temperature: 35
```

however if we look at the entity:

image

piitaya:
Hum, may be HA doesn't expose the min/max attributes and just use it to compute the state 😢

greghesp:
> Hum, may be HA doesn't expose the min/max attributes and just use it to compute the state 😢

Hopefully not, otherwise it's somewhat pointless, although maybe thats why the patched component is so heavily used?  I've raised a bug for it regardless: https://github.com/home-assistant/core/issues/70583

greghesp:
Updated so that no min/max values are provided through the card itself.  Instead, they need to be added through the customize section of `configuration.yaml`: https://github.com/greghesp/lovelace-mushroom/blob/plant-card/.hass_dev/configuration.yaml#L10

greghesp:
Should be ready for review now :)

piitaya:
Hi 🙂 I prefer to have the attributes exposed by home assistant before merging it. I will it if it's a big deal to add that 🙂

piitaya:
I'm speaking about the `limits` fields. I'm not sure it will be under `limits`. I think we can directly add the `min` `max` attributes at the root level.

piitaya:
Also, there is hardcoded hexa code in the `css`. It can be good to use css variable to support theme. But I can do this later if you're not familiar with that 🙂

piitaya:
There is also conflits with the `colors.ts file`. Entity related colors moved to `theme.ts`.

piitaya:
It seems that the goal of HA is not to maintain this type of entity because it's not a real entity, it's just a helper for the front-end. I'm not sure we should support that in Mushroom.

chrisb86:
@ greghesp Is it possible to release the card as a seperate repo outside of the mushroom project? I really like the design and would love to implement it in my dashboard. Are there any technical constraints?

rononline:
Will this be merged? I love to use the plant card. @greghesp or is it possible to use it without it being merged?

cromelex:
Would love to get this even if it has to be done separately outside of the core mushroom.

quentinus95:
@piitaya what is your final decision on this? It seems perfectly fine to have this card as long as Plant Monitor is kept in Home Assistant, and remove it if they decide to remove the helper. What do you think?

rononline:
> 

Agree on that! Could you PLEASE consider it to put this card in?

PR #422, opened 2022-04-30
fix: improve default responsiveness for actions

piitaya:
## Description
The default `double_tap_action` slow down the `tap_action`. Removing them by default improve responsiveness 🚀


bramstroker:
Nice one, was setting this to none each time to improve reactivity. Glad it's the default option now.

Issue #505, opened 2022-05-18
[Feature]: Stack card


acesyde:
### Is your feature request related to a problem?

Maintained support of stack card in mushroom

### Describe the solution you'd like

Reuse the amazing works from [Vertical stack in card](https://github.com/ofekashery/vertical-stack-in-card) and [Stack in card](https://github.com/custom-cards/stack-in-card) with integrated editor

PR #506, opened 2022-05-18
feature: stack card

acesyde:
## Description
- Add maintained version of stack in card / vertical stack in card in Mushroom

## Related Issue
#505

## Motivation and Context
- Previous mentionned repositories are not maintained 

## How Has This Been Tested
 Locally

## Types of changes

## Checklist

### DIFF
--- a/src/cards/stack-card/const.ts
+++ b/src/cards/stack-card/const.ts
@@ -1,0 +1,4 @@
+import { PREFIX_NAME } from "../../const";
+
+export const STACK_CARD_NAME = `${PREFIX_NAME}-stack-card`;
+export const STACK_CARD_EDITOR_NAME = `${STACK_CARD_NAME}-editor`;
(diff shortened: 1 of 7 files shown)


piitaya:
If we import this feature in mushroom (I'm not sure about it). I don't think we must import code from others repositories. I think it's better to improve other repo.

acesyde:
@piitaya for sure, but other repositories are not maintained :(

piitaya:
Maybe we can create a new repository for that?

piitaya:
I don't think I will have time to maintain this card so we will have the same issue as other repositories 🤣

alessandroias:
I think this is a great idea: `stack-in-card` does not support a visual editor making the mushroom's efforts useless.
image
This card does not require much maintenance, but will allow, in the future, to build custom buttons (aesthetically similar to those of the `vacuum`,  `fan` or `media` card) that integrate with other cards and perform actions, for example you could attach buttons to a fan entity to set the various modes. I realize it's a rather advanced use, difficult to maintain and perhaps even rare, but it would solve many problems of the vacuum and fan cards.

piitaya:
Vertical stack in card has a community. I prefer to help the developer to migrate to editor than copying his/her work.
Also I don't want users to install Mushroom only for the vertical stack in card included.

Issue #303, opened 2022-04-14
Add a "name & state" option for the content of the entity chips cards


cromelex:
I want to use a few chip cards showing the battery % for a couple of entities. Using a chip entity card, I can only display the name _or_ the percentage, not both at the same time. If  I have 2 different entities, I can either look at the icon (which is dynamic for battery) and show the  state, or only have the stage populated and not know exactly what entity it relates to.

I would like to be able to select a "name & state" for the content dropdown. This could just display the 2 items like "name"&"state", ie:
Name: Device1 and State: 50% to display as "Device1 50%". By keeping the name field editable, people could add a separator if needed, ie "Device1 :" to display "Device1 : 50%".

I have considered using the template chip, but the icon for the battery entities updates at every 10%, ie it is dynamic.
To avoid losing that feature, I would have to add all those steps to the template.

Thanks

joaovitoriasilva:
I would also love this implemented. Thanks

jlbphotos:
yes please, great idea 👍 )

PR #527, opened 2022-05-26
Add a "name & state" option for the content of the entity chips cards

bergander:
## Description

Add a "name & state" option for the content of the entity chips cards according to enhancement request #303
## Related Issue

#303
## How Has This Been Tested

Tested locally by adding an entity chip with the new state

## Types of changes

## Checklist

### DIFF
--- a/docs/cards/entity.md
+++ b/docs/cards/entity.md
@@ -11,18 +11,18 @@ A entity card allow you to display an entity.
 
 All the options are available in the lovelace editor but you can use `yaml` if you want.
 
-| Name                 | Type                                                | Default     | Description                                                                         |
-| :------------------- | :-------------------------------------------------- | :---------- | :---------------------------------------------------------------------------------- |
-| `entity`             | string                                              | Required    | Entity                                                                              |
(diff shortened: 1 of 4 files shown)


piitaya:

Template chip allow you to configure the information you want. I don't think we should provide this option as people will request name + state, name + last updated, etc... so I prefer to close the PR.

As alternative, you can use HA badge that support that option : https://www.home-assistant.io/dashboards/badges/
Or the template chip that allow you to have anything you want as state.

wagnst:
disagree, badges are not configurable within the dashboard itself, hence i'd love to use the chips card. But it needs to be much more configurable!! Just imagine starting scene entities which dont really have a state. Shows `unknown` for me.

piitaya:
You can display the name on chips and on badge.
I don't see how this PR can help your issue about scene. This PR adds an option to add state after the scene so you will have `scene name - unknown` which is not useful.

For future update, I prefer to put effort on the possibility to add badges everywhere in the dashboard instead of trying to copy the badges options to chips. In the end, when we will be able to place badges everywhere, chips will become deprecated as they don't have any value left.

Issue #576, opened 2022-06-26
[Bug]: Slider doesn't "snap" to step values.


acshef:
### Current Behavior

Example, given a fan with step size 33.3333:

1. Fan is initially off
2. User drags or tap position ~40%
3. Slider "snaps" to position 33% (fan is set to 33%, aka low) - *this is expected, existing behavior in mushroom*
4. User again drags or taps again to position ~40%, slider settles at 40% position instead of moving to nearest step

### Expected Behavior

Upon selecting a different *value* that ultimately resolves to the same *step position*, the slider should "snap back" to the original step position.

### Steps To Reproduce

See example above (use a fan entity with step 33.3333)

### Environment

```markdown
- OS:Ubuntu 20.04
- Browser: chrome
- Version: 1.10.0
- Hassio: 2022.6.7
```

### Anything else?

I have a PR ready to address this 😁 https://github.com/piitaya/lovelace-mushroom/pull/582

acshef:
Holy cow, congrats on v2.0.0! I upgraded to mushroom 2, and confirmed the issue is still present. (I'll rebase/revise my PR in the future, once this ticket is deemed worthy)

KTibow:
@acshef if #582 fixed this issue, can it be closed?

acshef:
Not sure of the preferences of this project; should issues be closed when a bugfix/feature PR is merged, or wait until there's a tagged release that contains the bugfix/feature?

piitaya:
"Fixes ###" auto close this issues. But only the issue number was present in the PR.
I close it manually 🙂

PR #582, opened 2022-06-27
Prevent selecting values that don't lie on a step

acshef:
## Description
This prevents from the input "settling" on a value that doesn't lie on a step.

## Related Issue
https://github.com/piitaya/lovelace-mushroom/issues/576

## Motivation and Context
In my opinion, I don't believe the slider should show a value that is not represented by the entity. Going from 33% to 40% doesn't speed up the fan -- the value of 33% is sent to Home Assistant both times.

This behavior is most visible on fans, and admittedly I don't have any other entities that I can test this with. But I assume the same expectation exists for all stepped entities (fans, covers, etc.)

## How Has This Been Tested
Given a fan with step size 33.3333:

1. Fan is initially off
2. User drags or tap position ~40%
3. Slider "snaps" to position 33% (fan is set to 33%, aka low) - *this is expected, existing behavior in mushroom*
4. User again drags or taps again to position ~40%, slider will not "snap back" to the 33% mark - *this PR aims to fix that behavior*

## Types of changes

- [X] 🐛 Bug fix (non-breaking change which fixes an issue)

## Checklist

- [X] My code follows the code style of this project.
- [X] I have tested the change locally.

### DIFF
--- a/src/shared/slider.ts
+++ b/src/shared/slider.ts
@@ -113,7 +113,8 @@ export class SliderItem extends LitElement {
                 if (this.disabled) return;
                 this.controlled = false;
                 const percentage = getPercentageFromEvent(e);
-                this.value = this.percentageToValue(percentage);
+                // Prevent from input releasing on a value that doesn't lie on a step
+                this.value = Math.round(this.percentageToValue(percentage) / this.step) * this.step;
                 this.dispatchEvent(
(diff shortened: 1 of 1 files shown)


piitaya:
With this change, we don't have a smooth slider isn't it ?

piitaya:
Ok, seems good to me 🙂

PR #588, opened 2022-07-01
Support charge in vacuum card

### LINKED ISSUE #456: [Feature]: Vacuum cleaner charge level
### Is your feature request related to a problem?

The vacuum cleaner card doesn't have an option to show the charge level of the robot vacuum cleaner.
afbeelding

### Describe the solution you'd like

Much like the media card has an option for showing the current volume, I'd suggest a similar option for the vacuum card to show the current charge.

### Describe alternatives you've considered

afbeelding

I've been using the vacuum card so far but it's too big and clunky for my purposes. It does however, have the charge level indicator. A similar thing could be applied to the mushroom vacuum card.

### Additional context

Thank you for your work and the great cards for home assistant! Your next coffee is on me 👍

KTibow:
Untested; May or may not fix #456


src/cards/vacuum-card/vacuum-card.ts
```diff
@@ -68,6 +68,33 @@ export class VacuumCard extends MushroomBaseCard implements LovelaceCard {
...
+
+    updateCharge() {
+        this.charge = undefined;
+        if (!this._config || !this.hass || !this._config.entity) return;
+
+        const entity_id = this._config.entity;
+        const entity = this.hass.states[entity_id] as VacuumEntity;
+
+        if (!entity) return;
+        this.charge = getCharge(entity);
```
piitaya, line 89: This can be directly in the render function. No need state for that.

piitaya (commented):
I don't know if you want this feature or not. Does the charge really important? I have this with my vacuum and I don't use this feature because it's always charged.

src/cards/vacuum-card/vacuum-card.ts
KTibow: I'm a bit new to the code in Mushroom, would you mind explaining what you mean by this?

KTibow: > I have this with my vacuum and I don't use this feature because it's always charged.

What type of vacuum do you use? When does it usually run, and for how long?

piitaya: Roborock s5 max. It take 1h15 for my first floor but it has 3h of battery so I don't care if the battery percentage 😅
May be we can add option to allow battery display or not.

KTibow: Sure. But what do you mean by

> No need state for that.

I was just copying the way it was implemented in all the other cards.

piitaya: We need state for light brightness and cover because there is a temporary value sent by the slider. 
The `onCurrentCharge` is linked to nothing so using a state for charge variable is not useful. We can just do this in the render function :

```ts
const charge = getCharge(entity);
```

Otherwise, I do not prefer to merge this PR for now because I prefer to find a way to allow user to choose what they want to display (a common way for all cards).

PR #599, opened 2022-07-08
Only have cursor: pointer when actions are applicable

### LINKED ISSUE #591: [Bug]: Cursor is pointer when all actions are disabled
### Current Behavior

Right now, if you disable all actions
image
The cursor is still a pointer
image

### Expected Behavior

The cursor would be normal when nothing happens upon click

### Steps To Reproduce

1. Make a template card
2. Set all actions to no action
3. Hover over card

### Environment

```markdown
- OS: Fedora Linux 36 (Workstation Edition)
- ~~Node:~~ *Browser*: Firefox 101.0.1
- *Mushroom* Version: 2.0.0
- ~~Hassio:~~ *Home Assistant*: 2022.6.7
```

### Anything else?

_No response_

KTibow:
Doesn't work yet in my testing, I'll work on debugging it - fixes #591

### DIFF
--- a/src/cards/cover-card/cover-card.ts
+++ b/src/cards/cover-card/cover-card.ts
@@ -253,9 +253,11 @@ export class CoverCard extends MushroomBaseCard implements LovelaceCard {
             super.styles,
             cardStyle,
             css`
-                mushroom-state-item {
-                    cursor: pointer;
-                }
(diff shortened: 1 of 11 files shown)


KTibow:
Does not work as intended, the inner part is still a pointer and the outer part of the card doesn't work.

PR #646, opened 2022-07-21
Arabic language translation

Hobiano:
## Description
Arabic support for Mushroom

## Types of changes

## Checklist


piitaya:
The `json` import is missing in the i18n file. I will add it.

Issue #544, opened 2022-06-06
[Feature]: add support for covers using tilt to cover card


Pocke01:
### Is your feature request related to a problem?

I have a roof on my pergola that only uses open_cover_tilt, close_cover_tilt, stop_cover_tilt and set_cover_tilt_position. 

Would love to be able to use the cover card with my roof. 

### Describe the solution you'd like

Maybe add a checkbox for covers only using tilt.

### Additional context

Love your work!  Finally easy to get a nice looking ui!

KTibow:
Possible duplicate of #434

Issue #150, opened 2022-02-21
Add support for current_cover_tilt_position


martindybal:
**Is your feature request related to a problem? Please describe.**
I really like the component, but it doesn't allow me to adjust the tilt. 

**Describe the solution you'd like**
Add a tilt setting option if the entity supports `SUPPORT_SET_TILT_POSITION`. Functioning and design could be similar to `show_position_control`. When the button is clicked, the tilt setting slider is shown. 

**Describe alternatives you've considered**
There could be a whole new UI, but it's probably too much complicated and big
![image](https://user-images.githubusercontent.com/7644099/154895734-ebbc835a-8955-436a-b192-4670fcc88337.png)
![image](https://user-images.githubusercontent.com/7644099/154895766-fe75d8d1-dccc-48ae-af9d-124ac448e319.png)
![image](https://user-images.githubusercontent.com/7644099/154895788-6b7ef4c6-33bf-4bd8-a530-10aaf5945e40.png)

martindybal:
Currently I'm using this setup. But it's not such a nice UI as `mushroom` provides
```
- type: entities
  entities:
    - type: custom:mushroom-cover-card
      entity: cover.103_office_blind_door
      show_buttons_control: true
      show_position_control: true
    - type: custom:slider-entity-row
      entity: cover.103_office_blind_door
      name: Tilt
      attribute: tilt
```
![image](https://user-images.githubusercontent.com/7644099/154896764-f398830c-8909-46e6-983b-c8075008c31e.png)

yolani:
+1 for this, all my 14 covers have a tilt position and I change the tilt level quite often on sunny days

kmplngj:
+1 form me. Need this 💯

martindybal:
My proposal:
- Allow customize buttons (custom icon and action)
- Add support for tilt position, similar UI as `show_position_control`

For example for me is ideal to have different buttons for each type of shades.
- For a blind I’m using fully open, sun protection (tilt adjusted to sun), fully closed. Custom level and tilt adjustment
- For roller shutter I’m using fully open, fully close, semi open (80%)

@piitaya can I help you with implementations?

redbear-ger:
any news on this one?

PR #739, opened 2022-09-22
Add support for tilt position slider in `CoverCard`

Atokulus:
## Description
This PR will add an option to control the tilt position within the cover card.

## Related Issue
- #150
- #544

## Motivation and Context
Over 80% of blind setups in Switzerland are tiltable. As tilt is equally important as position with these setups, support for tilt position control is highly recommended and asked for.

## How Has This Been Tested
This PR has been tested with the demo platform, both with `cover.living_room_window` (with working position and tilt) as well as `cover.hall_window` (which grays out the tilt controls correctly).

The newly added tilt control uses a linear gradient pattern. This lets the user better distinguish between both control modes.

The following things have been checked:

- Latest stable versions of Chrome and Edge
- State and value updates
- Config UI
- Transitions
- Light & Dark Theme
- Control Indicator
- Multiple curved sliders with different sizes on one dashboard

image

## Types of changes

## Checklist


piitaya:
Thanks for the PR.
I think we should find an other way to show the difference between tilt and position slider. It looks buggy to have a curved one.

Atokulus:
Dear @piitaya 
Thanks for checking it out. I agree with you, that the wide curved slider might be too similar to the regular one, and the hard corners are quite nasty looking. Yet I believe we should stick with the curvature concept:
1. The curvature resembles and reflects the tilt rotation in a quite nice way, so the user can remember/guess the meaning of the tilt control.
2. The slider is quite large so the user has fine control over the tilt, yet it integrates nicely with the existing patterns (no complicate or hidden controls for ease of access).

Therefore I have already implemented another proposal using a thinner, rounded SVG path. I guess this looks more promising.

image

Best regards
Markus

piitaya:
I think we must keep the current slider design but may be change the background like color slider or color temp slider.

Atokulus:
Dear @piitaya 
Maybe some line pattern that gets denser to 100% would do?
Best regards
Markus

Atokulus:
Dear @piitaya
This is my new proposal:

image

Best regards
Markus

src/cards/cover-card/controls/cover-tilt-position-control.ts
```diff
@@ -0,0 +1,78 @@
...
+        if (i !== 0) {
+            gradient.push([stopOffset1, 'transparent']);
+        }
+        gradient.push([stopOffset1, 'var(--slider-bg-color)']);
+        gradient.push([stopOffset2, 'var(--slider-bg-color)']);
+        gradient.push([stopOffset2, 'transparent']);
+    }
+
+    return gradient;
+}
```
piitaya, line 24: It must be computed one time instead of having a function called at each render.

src/cards/cover-card/cover-card.ts
```diff
@@ -121,6 +127,9 @@ export class CoverCard extends MushroomBaseCard implements LovelaceCard {
     @state()
     private position?: number;
 
+    @state()
+    private tilt_position?: number;
+
```
piitaya, line 132: It's not needed as the tilt position is not displayed in the state.

src/cards/cover-card/cover-card.ts
```diff
@@ -229,20 +255,36 @@ export class CoverCard extends MushroomBaseCard implements LovelaceCard {
...
+                        <mushroom-cover-position-control
+                            .hass=${this.hass}
+                            .entity=${entity}
+                            @current-change=${this.onCurrentPositionChange}
+                            style=${styleMap(sliderStyle)}
+                        />
+                    `;
+                }
+            case "tilt_position_control": {
+                const color = getTiltStateColor(entity as CoverEntity);
```
piitaya, line 274:
```suggestion
                const color = getStateColor(entity as CoverEntity);
```
The same color should be used for the control. No need to have a separate function and new css variables.

src/shared/curved-slider.ts
```diff
@@ -0,0 +1,376 @@
+import { css, CSSResultGroup, html, svg, LitElement, PropertyValues, TemplateResult } from "lit";
```
piitaya, line 1: This file can be deleted

src/utils/theme.ts
```diff
@@ -53,6 +53,9 @@ export const themeVariables = css`
     /* Slider */
     --slider-threshold: var(--mush-slider-threshold);
 
+    /* Curved Slider */
+    --control-curved-slider-thickness: var(--mush-control-curved-slider-thickness, 12px);
+
```
piitaya, line 58: Can be deleted

src/utils/theme.ts
```diff
@@ -136,6 +139,10 @@ export const themeColorCss = css`
     --rgb-state-cover-open: var(--mush-rgb-state-cover-open, var(--rgb-blue));
     --rgb-state-cover-closed: var(--mush-rgb-state-cover-closed, var(--rgb-disabled));
 
+    /* State cover tilt colors */
+    --rgb-state-cover-tilt-open: var(--mush-rgb-state-cover-tilt-open, var(--rgb-blue));
+    --rgb-state-cover-tilt-closed: var(--mush-rgb-state-cover-tilt-closed, var(--rgb-disabled));
+
```
piitaya, line 145: Can be deleted

piitaya (changes requested):

src/cards/cover-card/controls/cover-tilt-position-control.ts
Atokulus: Ok. Was assuming as it is in `static get styles(): CSSResultGroup;`, this would be evaluated only once nonetheless.

Atokulus: No problem ;)

src/cards/cover-card/cover-card.ts
Atokulus: I removed `tiltStateColor`. Let's see how people use it, and maybe there isn't any difference in perception ;)

src/cards/cover-card/cover-card.ts
Atokulus: Removed

src/cards/cover-card/controls/cover-tilt-position-control.ts
Atokulus: Evaluate GRADIENT as const nonetheless

src/shared/curved-slider.ts
Atokulus: Deleted

src/utils/theme.ts
Atokulus: Deleted

src/utils/theme.ts
Atokulus: Deleted

acseven:
Thanks for this addition, but would it be feasible to control the tilt angle without having a specific blinds tilt control (for example Shelly 2.5) and maybe using a time span for calculation of the blinds position?

piitaya:
@zylxpl We don't have this information so we can adapt the UI 😞

@Atokulus Is your cover open at 0 and closed at 100?

zylxpl:
@piitaya What information? This should be user configurable, simple switch to alter the slider to smth like this:
image

PR #759, opened 2022-10-03
Allow using 'source' for app icon

### LINKED ISSUE #758: [Bug]: Media card doesn't use the 'source' attribute for it's media info
### Current Behavior

When settings a media card to read from a TV entity and the 'Use media info' option is set to on, only the default TV icon will show if the integration uses a 'source' attribute to inform Home Assistant what is being played on the device.

### Expected Behavior

When settings a media card to read from a TV entity and the 'Use media info' option is set to on, the icon for the relevant app will be displayed

### Steps To Reproduce

1. Have an entity which uses 'Source' to show what is set on the TV (such as from the webOS integration)
2. Create a 'media' card and enable 'Use media info'

### Environment

```markdown
- OS: HASS OS
- Browser: Chrome/ Safari/ Firefox
- Version: 2.1.3
- Hassio: 2022.9.7
```

### Anything else?

Here's an example of the state from my TV from the webOS (LG) integration:

```yml
source_list:
  - Apple TV
  - Disney+
  - ITV Hub
  - My5
  - NOW
  - Netflix
  - PS5
  - Plex
  - Spotify - Music and Podcasts
  - Twitch
  - UKTV Play
  - YouTube
volume_level: 0.12
is_volume_muted: false
source: PS5
sound_output: tv_speaker
assumed_state: true
device_class: tv
friendly_name: Living Room TV
supported_features: 24509
```

JeffersonBledsoe:
## Description
This PR allows integrations which use `source` rather than `app_name` to be given an app icon

## Related Issue

Closes #758 

## Motivation and Context

Users may not want to use the full app picture but instead have an icon that fits in with the HA UI so they can use the icon. Some integrations didn't provide an `app_name` attribute for this, but instead uses `source`

## How Has This Been Tested

Locally using the instructions in README.md to develop against HA.

## Types of changes

## Checklist

### DIFF
--- a/src/cards/media-player-card/utils.ts
+++ b/src/cards/media-player-card/utils.ts
@@ -64,29 +64,41 @@ export function getVolumeLevel(entity: MediaPlayerEntity) {
         : undefined;
 }
 
+const mediaIconMapping = {
+    app_name: {
+        spotify: "mdi:spotify",
+        "google podcasts": "mdi:google-podcast",
+        plex: "mdi:plex",
+        soundcloud: "mdi:soundcloud",
+        youtube: "mdi:youtube",
+        "oto music": "mdi:music-circle",
+        netflix: "mdi:netflix",
+    },
+    source: {
+        "google podcasts": "mdi:google-podcast",
+        "oto music": "mdi:music-circle",
+        netflix: "mdi:netflix",
+        ps5: "mdi:sony-playstation",
+        ps4: "mdi:sony-playstation",
+        playstation: "mdi:sony-playstation",
+        plex: "mdi:plex",
+        soundcloud: "mdi:soundcloud",
+        spotify: "mdi:spotify",
+        twitch: "mdi:twitch",
+        youtube: "mdi:youtube",
+    },
+};
+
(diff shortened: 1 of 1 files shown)


src/cards/media-player-card/utils.ts
```diff
@@ -64,29 +64,41 @@ export function getVolumeLevel(entity: MediaPlayerEntity) {
...
-                return "mdi:netflix";
-            default:
-                return icon;
-        }
+        Object.keys(mediaIconMapping).forEach((attribute) => {
+            var app = entity.attributes[attribute]?.toLowerCase();
+            if (app) {
+                icon = mediaIconMapping[attribute][app];
+            }
+        });
```
piitaya, line 101: It's clear about the rules priority because it depends of the order of the attributes.
I think we can do like this (I just wrote a pseudo code but you got the idea 🙂)
```
for each app_name
    if app_name === entity.attribute.app_name:
           return app_name icon
for each source
    if app_name === entity.attribute.source:
           return source icon
```

piitaya (changes requested):

Issue #800, opened 2022-10-27
[Feature]: Custom Off States for Entities


murkumurdok:
### Is your feature request related to a problem?

Some media player devices have more "off" states instead of on/off.  For example, my AppleTV has the state of "standby" when it is actually off or sleeping.  Because of this, the mushroom card still colors the icon for the AppleTV as if the device was still on.

### Describe the solution you'd like

Include a configuration option for entity/media mushroom cards that allow you to set custom "off" states per entity so that the card will properly appear on or off.

piitaya:
I think it's better to consider standby as inactive for every media player.

murkumurdok:
> I think it's better to consider standby as inactive for every media player.

Sure, that works too

murkumurdok:
Thank you

murkumurdok:
Hello, my AppleTV device is still showing as turned on when in standby on the mushroom media card.

murkumurdok:
No worries, it happens, thank you!

PR #802, opened 2022-10-29
Improve script, button, scene and media player active state.

piitaya:
## Description

Always consider script/scene/button/input button as active.
Consider standby and media player as inactive.

## Related Issue
Fixes https://github.com/piitaya/lovelace-mushroom/issues/800

## Types of changes

## Checklist

- [ ] I have tested the change locally.

### DIFF
--- a/src/ha/data/entity.ts
+++ b/src/ha/data/entity.ts
@@ -1,4 +1,5 @@
 import { HassEntity } from "home-assistant-js-websocket";
+import { computeDomain } from "../common/entity/compute-domain";
 
 export const UNAVAILABLE = "unavailable";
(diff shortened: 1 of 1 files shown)


alexandrechoske:
@piitaya amazin job on cards, may I suggest something?
This missed function made me use Button card instead of mushroom, because on button-card I could do that.
I missed some "responsive" button press, like we can do on button-card with this css

```
      styles:
        card:
          - '--mdc-ripple-color': yellow
          - '--mdc-ripple-press-opacity': 0.5
```

So when I press the button it looks like an "wave" yellow effect.
image

If you add this as a option if the user want to use that, would be awesome!
Thank you!

PR #848, opened 2022-11-12
Set min light brightness to 1

### LINKED ISSUE #847: [Bug]: Swiping a light down to the lowest brightness turns it off
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue.

### Current Behavior

When using the light card and using the brightness slider, there is no easy way to set the light to the lowest brightness. Swiping the brightness all the way to the left (to 0%) turns the light off, which can already be accomplished by simply clicking the card.

### Expected Behavior

Swiping the brightness down sets the brightness to 1%, and does not turn off the light.

### Steps To Reproduce

1. Create a light card with brightness control
2. Turn the light on
3. Swipe down the brightness all the way to the left

### Context

_No response_

### Environment

```markdown
- Browser: Firefox
- HA Version: 2022.11.2
- Mushroom: 2.3.6
```

### Anything else?

_No response_

wilco375:
## Description
This PR updates the light brightness slider such that it goes down to 1% instead of 0%. See related issue for reasoning.

## Related Issue

This PR fixes or closes issue: fixes #847

## Motivation and Context

Allows setting a light to its lowest brightness without turning it off

## How Has This Been Tested

New code has been compiled and added to HA as described in the README under _Manual_

## Types of changes

- [x] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [ ] I followed [the steps](https://github.com/piitaya/lovelace-mushroom#maintainer-steps-to-add-a-new-language
) if I add a new language .


wilco375:
This could also be implemented as an extra switch in the light card to choose between 0% or 1% as a minimum if that is more desirable.

piitaya:
I think it must `1` but we need to make the icon clickable before to have a easy way to turn off the light. It can be confusing to have it as an option.

wilco375:
> I think it must `1` but we need to make the icon clickable before to have a easy way to turn off the light. It can be confusing to have it as an option.

I'm not sure I follow what you mean; the icon is already clickable, right?

antonio1475:
> Maybe I'm misunderstanding you, but changing the tap action is already possible in Mushroom, right? 
> 

Mushroom has tap actions for the whole card (outside of the slider)
The new Tile Card splits the tap_action between the whole card and the icon:
image
(one action for the icon in yellow, one action for the rest of the card in green)

In any case, @piitaya I don't really see the dependency to that, since mushroom already has even more (3) actions (Tap, Hold, Double Tap), so users can still configure any of those 3 to toggle (which is the default anyway). No?

> Or is your point that some users might have changed the tap action and use swipe to turn off the light?
That is of course a risk, but it could be announced as a Breaking change and users can re-configure any of the 3 actions to toggle?

wilco375:
> Mushroom has tap actions for the whole card (outside of the slider) The new Tile Card splits the tap_action between the whole card and the icon

Ah, thanks for the clarification

samuolis:
@piitaya Is it hard to add support for mushroom cards to allow icon click? Maybe you have timeline when it will be available in mushroom?

krimo80:
When will this fix be applied to the release..? ;-)

Snellingen:
What is blocking this fix right now? Is it that this change has have an option for the card to set min/max for the slider to avoid breaking changes?

piitaya:
Let's merge it to be consistent with tile card 🎉

NegussugeN:
Apologize if I am completely missing the boat here, but I just updated to 4.3.0 and even after fully rebooting the system running HA my brightness sliders still go down to 0% instead of 1%. Is there something else I need to do or an I misunderstanding the purpose of this PR? Much appreciated!

Issue #664, opened 2022-07-27
Climate Temperature control increaments


mtgpcs:
### Current Behavior

Currently When changing the temperature up or down it goes in 0.5 increments.

### Expected Behavior

I would like to be able to set the increment amount my AC only does whole amounts. 

### Steps To Reproduce

If you go down 26  it goes to 25.5 or if you go up  it goes 26 to 26.5.

### Environment

```markdown
- OS: Home Assistant 2022.7.7
- Node:  Supervisor 2022.07.0
- Version: Operating System 8.4
- Hassio: Frontend 20220707.1 - latest

I just installed mushrooms today.
```

piitaya:
Hi 🙂 Each climate entity expose `target_temp_step` (and the default is 0.5 in Home Assisant).
You can customize this in your `configuration.yaml` if you want :

https://www.home-assistant.io/docs/configuration/customizing-devices/#manual-customization

```yaml
homeassistant:
  customize:
    thermostat.family_room:
      target_temp_step: 1
```

antonio1475:
> Hi 🙂 Each climate entity expose `target_temp_step` (and the default is 0.5 in Home Assisant). You can customize this in your `configuration.yaml` if you want :
> 
> https://www.home-assistant.io/docs/configuration/customizing-devices/#manual-customization
> 
> ```yaml
> homeassistant:
>   customize:
>     thermostat.family_room:
>       target_temp_step: 1
> ```

I've been researching this for weeks. Multiple people asking the same. Nobody had an answer.
I came to request it for the Mushroom card, searched beforehand and there you are, solving the question at a system level. You're the man! Thank you 🙏

piitaya:
It's always better to fix the thing at the system level instead of providing an option because it will be fixed in every card 👍

jejegri:
Hi piitaya,

I have tried adding these line to configuration.yaml file and it works... but only for the + button. For -, it is still going by 0,5 step. Any idea how to resolve this issue ?
This is using Tado thermostat via Homekit.
Thanks :) 

Méthode d'installation : Home Assistant OS
Core : 2026.2.3
Supervisor : 2026.02.2
Operating System : 17.1
Interface utilisateur : 20260128.6

PR #932, opened 2023-01-06
Temperature step size +/- 1 degree

tomg1970:
Hello,
it would be great if you could define for the climate-card the step size of the
temperature setting could be defined. 
Because my zigbee MOES thermostat can eb only 1.0 degree steps 
not +/- 0.5 degrees.
The card would be perfect, if you could define a temperature min and max range.
could be defined.


piitaya:
Hello. There was already an issue for this : https://github.com/piitaya/lovelace-mushroom/issues/664

Issue #934, opened 2023-01-08
[Feature]: More precise slider movements for touch input


phischdev:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

It is hard to do fine-grained changes to sliders by touch on a small device like a smartphone.
Hitting exactly 1% (instead of 0%) or other small values requires many tries.

While during day-time a difference of few percentages points aren't noticeable, at night in a low-light setting, the difference between 1% or 4% brightness of a main light source makes a strong perceivable difference. 
Even more so, choosing a color from the color band is difficult to do precisely.

### Describe the solution you'd like

Instead of having to hit the absolute value by finger (like the left-most pixel of the slider for 0%), the slider could be moved relatively at lower speed. 

Example: the slider is at 10%. The finger touches down anywhere on the right half of the slider and moves to towards the left. The value is then decreased by half the way the finger travelled.

This would allow for precise placement of the slider. Similar to how scrubbing works in many smartphone apps like the audible player.

### Describe alternatives you've considered

Instead of the slide speed being fixed at eg. 50% it could be adjusted by dragging the finger downwards like many other apps do.

### Additional context

I hope the description of the problem and the solution are understandable. 

I already implemented this solution and was about to create a PR when I saw it requires creating an issue first.
It can be found here (https://github.com/piitaya/lovelace-mushroom/pull/935)

Notice: The changes only affect the panning (pointer down, move, pointer up). Clicking/Tapping an absolute value still works like before.

piitaya:
I don't think we should add this type of control. If you move your finger from 30px, the slider should move by 30px. 

The proper way to have more precise control is to use a full screen slider like this one :
![CleanShot 2023-01-09 at 17 08 25](https://user-images.githubusercontent.com/5878303/211354163-5c2fc054-a76d-484e-b462-9785db38fa2b.png)

milandzuris:
> I don't think we should add this type of control. If you move your finger from 30px, the slider should move by 30px.
> 
> The proper way to have more precise control is to use a full screen slider like this one : ![CleanShot 2023-01-09 at 17 08 25](https://user-images.githubusercontent.com/5878303/211354163-5c2fc054-a76d-484e-b462-9785db38fa2b.png)
When is release date for new HA cards and other dashboards things?

phischdev:
[...]

PR #935, opened 2023-01-08
More precise slider movement for pan gestures

phischdev:
## Description

Instead of having to hit the absolute value by finger (like the left-most pixel of the slider for 0%), the slider could be moved relatively at lower speed.

Example: the slider is at 10%. The finger touches down anywhere on the right half of the slider and moves to towards the left. The value is then decreased by half the way the finger travelled.

This would allow for a precise placement of the slider. Similar to how scrubbing works in many smartphone apps (like the audible player).

## Related Issue

This PR fixes or closes issue: https://github.com/piitaya/lovelace-mushroom/issues/934

## Motivation and Context
It is hard to do fine-grained changes to sliders by touch on a small device like a smartphone.
Hitting exactly 1% (instead of 0%) or other small values requires many tries.

While during day-time a difference of few percentages points aren't noticeable, at night in a low-light setting, the difference between 1% or 4% brightness of a main light source makes a strong perceivable difference.
Even more so, choosing a color from the color band is difficult to do precisely.

## How Has This Been Tested
I tried all edge cases of changing a slider value:
- overflowing values
- underflowing values 
- undefined values after switch was off
Directly clicking the slider to set an absolute value still works

## Types of changes

-   [X] ⚠️ Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

-   [X] My code follows the code style of this project.
-   [X] I have tested the change locally.


src/shared/slider.ts
```diff
@@ -73,6 +66,13 @@ export class SliderItem extends LitElement {
     private slider;
 
     setupListeners() {
+        const getPercentageFromEvent = (e: HammerInput) => {
+            const x = e.center.x;
+            const offset = e.target.getBoundingClientRect().left;
+            const total = e.target.clientWidth;
+            return (x - offset) / total;
```
phischdev, line 73: Moved clamping to another function, so that percentages in the negative scale or bigger than 100% can be received.

src/shared/slider.ts
```diff
@@ -87,10 +87,23 @@ export class SliderItem extends LitElement {
...
-            this._mc.on("panstart", () => {
+            
+            let panstartPercentage = 0;
+            const getPanTargetValue = (e) => {
+                const percentage = getPercentageFromEvent(e);
+
+                const deltaPercentage = (percentage - panstartPercentage) * 0.5;
+                const deltaValue = this.percentageToValue(deltaPercentage);
+                
+                return Math.max(Math.min(savedValue + deltaValue, this.max), this.min)
```
phischdev, line 98: Clamp value here

src/shared/slider.ts
```diff
@@ -87,10 +87,23 @@ export class SliderItem extends LitElement {
...
+            }
+
+            this._mc.on("panstart", (e) => {
                 if (this.disabled) return;
                 this.controlled = true;
-                savedValue = this.value;
+                savedValue = this.value || 0;
```
phischdev, line 104: default to `0` if `this.value` is undefined (light was off)

src/shared/slider.ts
```diff
@@ -87,10 +87,23 @@ export class SliderItem extends LitElement {
             this._mc.add(new Hammer.Tap({ event: "singletap" }));
 
             let savedValue;
-            this._mc.on("panstart", () => {
+            
+            let panstartPercentage = 0;
+            const getPanTargetValue = (e) => {
+                const percentage = getPercentageFromEvent(e);
+
+                const deltaPercentage = (percentage - panstartPercentage) * 0.5;
```
phischdev, line 95: slider moves 0.5 times the speed of the finger

PR #1066, opened 2023-03-11
Allow primary_info and secondary_info to be templates if "template" is selected

basnijholt:
Before I continue to implement this, I would like to know whether such a change is likely to get merged or not.

## Related Issue

This PR fixes or closes issue: fixes #

## Types of changes

## Checklist

-   [ ] My code follows the code style of this project.
-   [ ] I have tested the change locally.


piitaya:
I didn't added template support to any card for multiple reasons : 
- template are back-end computed so it can not be rendered so we can not use `computeInfoDisplay`. There is a subscribe/unsubscribe logic in template card for this.
- template can be very cpu intensive because it create lots of back-end request and backend computing.

For these reason, I preferred to keep it separated from others cards so only advanced users use it. I think we need front-end templating to improve performance and usage but it will not be as powerful as back-end ones.

Issue #888, opened 2022-12-09
[Feature]: Add fan presets


MartijnTuinstra:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

There currently is no support for presets in the fan-card

### Describe the solution you'd like

An option to select a preset out of a list of all available presets.

### Describe alternatives you've considered

Buttons for each preset requires a lot of configuration as:
- not all presets can be mapped to an icon
- maybe to many presets to show (and therefore might not fit)

### Additional context

My idea of how it could work

![image](https://user-images.githubusercontent.com/33627132/206806894-ce485396-b7d1-48c0-95bc-80ca12091e6b.png)

RickeyWard:
see: https://github.com/piitaya/lovelace-mushroom/issues/1032
![image](https://user-images.githubusercontent.com/342468/229974950-3f3f282b-0ba9-433a-bb71-23d911deb5b1.png)

RickeyWard:
I've got a PR open for this, but @piitaya isn't sure if he wants to include the feature. If he decides against it but you still want it , I created an addon build to be used with mushroom (not standalone but in addition to it) for those that find them selves here and are interested.

[Mushroom Fan Card Extended Gist](https://gist.github.com/RickeyWard/28408f3f210ad64a4ecaa2d936d35cfe)

jeroenvdd:
My fan has presets set, I would also like a way to show these presets in the gui.
the climate card has gotten this love trough the possibility to select HVAC modes,
![image](https://github.com/user-attachments/assets/eb4fadbd-c243-4852-94a8-62dad2550bdf)


it shouldn't be that hard to provide same functions for:
![image](https://github.com/user-attachments/assets/181ea9cf-9261-4404-8df3-c419809fdf1e)

Issue #1032, opened 2023-02-20
[Feature]: Preset levels for Lights, Fans, etc instead of slider


bobloadmire:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

I would like to have a set of buttons inplace of the slider, something like low/medium/high that I could assign values of something like 10/100/255 to. I don't need the granularity of the slider, buttons would be better.

### Describe the solution you'd like

I would like to have a set of buttons inplace of the slider, something like low/medium/high that I could assign values of something like 10/100/255 to. I don't need the granularity of the slider, buttons would be better.

### Describe alternatives you've considered

Theres another card called "brightness preset row" that works great, but it doesn't follow the Mushroom theme.

RickeyWard:
Working on this for fans, 
![image](https://user-images.githubusercontent.com/342468/229974164-7f0adc47-1f36-47e0-adea-264222d1870a.png)
![image](https://user-images.githubusercontent.com/342468/229974950-3f3f282b-0ba9-433a-bb71-23d911deb5b1.png)

Not sure how to a list editor into the config without a lot of extra work so defining presets is yaml but it still lets you edit the rest in visual editor.

bobloadmire:
That's perfect. Any way to easily modify it for lights? Should work the same way, just specify 3 values 0-255?

Luupw3d:
> Working on this for fans, ![image](https://user-images.githubusercontent.com/342468/229974164-7f0adc47-1f36-47e0-adea-264222d1870a.png) ![image](https://user-images.githubusercontent.com/342468/229974950-3f3f282b-0ba9-433a-bb71-23d911deb5b1.png)
> 
> Not sure how to a list editor into the config without a lot of extra work so defining presets is yaml but it still lets you edit the rest in visual editor.

It doesn't work.

<img width="922" height="734" alt="Image" src="https://github.com/user-attachments/assets/ba0cd1de-10a1-48e4-8b83-c6aef5db0ba6" />

RickeyWard:
> It doesn't work.

It was never merged, mushroom maintainers didn't like it. Said it was confusing.

If you want to appended it onto mushroom you can follow this https://gist.github.com/RickeyWard/28408f3f210ad64a4ecaa2d936d35cfe

It's not well maintained but it works good enough for my use cases

PR #1095, opened 2023-04-05
Fan Card Presets

RickeyWard:
## Description

Added the ability to add a row of presets (instead of or in addition to slider)

image
image

A decision could be made if the buttons in default/vertical mode should be on their own line or not. I prefer them on their own line but I don't use a fan that has oscillate controls. Having the oscillate control on its own line when the slider is hidden has always looked odd to me, so it looks extra odd now. but not different than it already looks.

image
image

Also note is that there doesn't seem to be any obvious way to do a "list" editor in the visual editor. home assistant has an entity-row editor but it seems purpose built not just generic for any scheme. Hopefully its not a deal braker that there's some optional yaml config, I realize that typically that isn't done here. I could write some custom editor I suppose.

## Related Issue

This PR closes issue: #888 part of #1032 and discussion #893

## Motivation and Context

Fan switches are often percentages in HA but in reality only honor a few values, and only a few values are usually valuable. Using a slider for a fan is awkward. Presets are reproducible especially on cheaper touchscreen devices.

## How Has This Been Tested

Running currently in my personal home assistant instance. Verified on ios 15, android with webview 110, chrome 111.0.5563.147 on windows 11.

## Types of changes

## Checklist

-   [x] I followed [the steps](https://github.com/piitaya/lovelace-mushroom#maintainer-steps-to-add-a-new-language) if I add a new language . (I didn't add a new  language)

### DIFF
--- a/docs/cards/fan.md
+++ b/docs/cards/fan.md
@@ -9,13 +9,15 @@ A fan card allows you to control a fan entity.
 
 ## Configuration variables
 
-All the options are available in the lovelace editor but you can use `yaml` if you want.
+Almost all the options are available in the lovelace editor but you can use `yaml` if you want.
+
 
 | Name                      | Type                                                | Default     | Description                                                                         |
 | :------------------------ | :-------------------------------------------------- | :---------- | :---------------------------------------------------------------------------------- |
 | `entity`                  | string                                              | Required    | Fan entity                                                                          |
 | `icon`                    | string                                              | Optional    | Custom icon                                                                         |
 | `name`                    | string                                              | Optional    | Custom name                                                                         |
+| `presets`                 | preset list                                         | Optional    | list of persentage presets and coresponding icons                                   |
 | `layout`                  | string                                              | Optional    | Layout of the card. Vertical, horizontal and default layout are supported           |
 | `fill_container`          | boolean                                             | `false`     | Fill container or not. Useful when card is in a grid, vertical or horizontal layout |
 | `primary_info`            | `name` `state` `last-changed` `last-updated` `none` | `name`      | Info to show as primary info                                                        |
@@ -28,3 +30,12 @@ All the options are available in the lovelace editor but you can use `yaml` if y
 | `tap_action`              | action                                              | `toggle`    | Home assistant action to perform on tap                                             |
 | `hold_action`             | action                                              | `more-info` | Home assistant action to perform on hold                                            |
 | `double_tap_action`       | action                                              | `more-info` | Home assistant action to perform on double_tap                                      |
+
+### presets
+
+presets are entered as a list of objects value and icon
+
+| Name                      | Type                                                | Default     | Description                                                                         |
+| :------------------------ | :-------------------------------------------------- | :---------- | :---------------------------------------------------------------------------------- |
+| `value`                   | integer                                             | Required    | 0-100 representing the percentage for the preset                                    |
+| `icon`                    | string                                              | Optional    | Custom icon (defaults to mdi:fan)                                                   |
--- a/docs/cards/fan.md
+++ b/docs/cards/fan.md
@@ -17,7 +17,7 @@ All the options are available in the lovelace editor but you can use `yaml` if y
 | `entity`                  | string                                              | Required    | Fan entity                                                                          |
 | `icon`                    | string                                              | Optional    | Custom icon                                                                         |
 | `name`                    | string                                              | Optional    | Custom name                                                                         |
-| `presets`                 | preset list                                         | Optional    | list of persentage presets and coresponding icons                                   |
+| `custom_presets`          | preset list                                         | Optional    | list of percentage presets and coresponding icons                                   |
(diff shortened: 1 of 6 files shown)


alessandroias:
I strongly believe that presets should not have their own line: leaving them as now generates visual inconsistency with the rest of the cards.

alessandroias:
I strongly believe that presets should not have their own line: leaving them as now generates visual inconsistency with the rest of the cards.

piitaya:
Hello 👋
Thanks for the contribution

Mushroom has always be UI first so we can not add this feature without editor support.
Also, presets already exist as an attribute for fan so we shouldn't name them presets. May be "shortcuts" ?
I strongly agree with @alessandroias. This new control should like any other, I should not be another line.

I have a mixed opinion on this feature with this new feature because the `2023.4` (see [beta releases notes](https://rc.home-assistant.io/blog/2023/03/29/release-20234/#fans)) will add this kind of buttons for fan in more info dialog and tile card if the fan have less than 3 speeds available. 
The UI is similar so it can be confusing to the user to have a different list of speed when it open the new dialog.

RickeyWard:
I'm excited to see builtin cards getting more love, but as far as I can tell from the diagnostic logs my embrigthten zwave fan switches do not report that they have preset_modes or what it's 3 fan speeds are, I'm guessing for compatibility its reporting that it has a value 0-99 as I doubt zwave-js is remapping it to be a percentage instead of fixed values so I'm assuming this new feature just won't work in the case of all my fan switches. If that's not the case, I'd like to update mushroom to honor the new feature not just ignore it.

As for the name, the attribute name is "preset_modes", I'd counter with "custom_presets" instead of "shortcuts" as its more semantically accurate but what appeases the rules is what I can go for, I'd like to get it merged so I don't  have to always run my own custom version and keep it upto date.

Putting them on the same line is no problem, Will do.

As for editing in the UI, Since the existing ha-form doesn't seem to support lists editing and that's pulled in from ha-frontend, would it be acceptable to break outside the ha-form and put in custom editing components? or would an acceptable stop gap be to use a string field field with a custom-presets format like `value|icon;value|icon;` ex `25|mdi:fan-speed-1;50|mdi:fan-speed-2;99|mdi:fan-speed-3` ?

piitaya:
I understand that you want do have custom speed in Mushroom but we can not add this until the speed presets are not configurable in the more info to be consistent across the full HA UI.

RickeyWard:
I don't quite understand what you're saying.

Multiple people have asked for it and it's related to 3 issues, I met all of requirements you brought up before.

What is this issue specifically?

piitaya:
Did you see the 2023.4 update ? https://www.home-assistant.io/blog/2023/04/05/release-20234/
image
image

It looks the same as you proposed (same layout, icons) because non-configurable.

RickeyWard:
I have this update and it does not do this for me, my fan switches are 3 position but show up in HA 2023.4 as percentages.

image
image

All my addons are up to date too, I went into zwave addon and "re-interviewed" my fan switches and its still a percentage. There's no configuration I can find in HA to change this.

cjsmns:
This card looks great!! I'd love to see them implemented. The consistency between the new stock tile cards and mushroom cards is a bit off, and placing them next to each other I can see a difference between the 2 and I don't like it. This would make it look more consistent.

Just to give some insight to the following comment:

> I have this update and it does not do this for me, my fan switches are 3 position but show up in HA 2023.4 as percentages.

I believe that your fan may not be reporting correctly, or something else odd is going on. With the latest up to 2023.4 it seems that only fans that support 3 speeds are supported. I have 2 fans that are 6 speeds and they only show the percentage slider, while my 3-speed fans show the new speeds options.

From what I can tell, to get the new options your fan needs to show it's percentage at 0%, 33%, 66% or 100%. Your fan is showing 25%.

RickeyWard:
> I believe that your fan may not be reporting correctly, or something else odd is going on.

I have interviewed my fan switch manually outside of HA and it just shows up as a percentage, 0 is off, [1, 32] is low, [33, 65] is med, [66, 99] is high. That's how it works and how its suppose to work.  Pressing up jumps up 33%, pressing down jumps down 33%, not to set increments just from where-ever it was. It's not a great fan switch but I've invested $150 in three and I don't want to change it just to get buttons on a webpage that I can make myself 🤷 

> From what I can tell, to get the new options your fan needs to show it's percentage at 0%, 33%, 66% or 100%. Your fan is showing 25%.

My fan shows its percentage as whatever I set it to, command class 38, mutlilevel-swtich min 0 max 99 😢 

I don't know if @piitaya is going to say anything else about it, but I'll leave this open until he closes it, for anyone interested you can use this [gist](https://gist.github.com/RickeyWard/28408f3f210ad64a4ecaa2d936d35cfe) to load this now along side mushroom

RickeyWard:
Clearly not interested. Closing.

ShayBox:
:disappointed: It looked so good too, a percent bar makes no sense for 3 speeds.

Issue #1189, opened 2023-06-30
[Feature]: Add markdown support to Title card


circa1665:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

Would it be possible to add basic Markdown support to the Title card?

It would be helpful if select text could be in bold or italicised etc.

### Describe the solution you'd like

<img width="497" alt="Screenshot 2023-06-30 at 20 37 26" src="https://github.com/piitaya/lovelace-mushroom/assets/33236082/a84df97c-d058-49d5-afa1-2af08092b360">

For example I have this welcome message on my main Dashboard that displays in the morning. It would be helpful if I could bold or italicise select words and phrases so that the key information is more easily distinguishable at a glance.

With markdown support I would be be able to style the text in my template using `**bold**` and `*italics*` like this:

```
{% if is_state("binary_sensor.is_it_morning", "on") %} {% set ignore = ["unknown", "unavailable"] %} {% if states("sensor.openweathermap_forecast_precipitation_probability") not in ignore and states("sensor.good_morning_weather_helper") not in ignore and states("sensor.good_morning_forecast_temp_helper") not in ignore %} {% set wet_weather = ("exceptional", "thunder and lightning", "rainy with thunder and lightning", "heavy rain", "rain showers", "snow and rain showers") %} {% set forecast_temp = states("sensor.good_morning_forecast_temp_helper") %} {% set weather_desc = states("sensor.good_morning_weather_helper") %} {% if states("sensor.good_morning_weather_helper") in wet_weather %}The weather today will be **{{ weather_desc }},** the probability of rain is {{ states("sensor.openweathermap_forecast_precipitation_probability") }}%, **don't forget an umbrella.**{% else %}The weather today will be **{{ weather_desc }}.**{% endif %} {% if is_number(forecast_temp) %}The forecast temperature is **{{ forecast_temp }}°C.**{% endif %} {% endif %}
  {% if is_state("binary_sensor.quote_of_the_day", "on") %}
{{ state_attr("binary_sensor.quote_of_the_day", "quotation") }}
*- {{ state_attr("binary_sensor.quote_of_the_day", "who_said_it") }}*
  {% endif %}
{% endif %}

Similar to the Markdown card: https://www.home-assistant.io/dashboards/markdown/


### Describe alternatives you've considered

_No response_

### Additional context

_No response_

[...]

Issue #767, opened 2022-10-11
[Feature]: HTML and markdown should be allowed in template


flemmingss:
### Is your feature request related to a problem?

markdown and HTML is possible to use in the normal markdown card
This makes it very easy to style how you want. But the template cards in this mushroom-cards collection does not support that.
Because of this I can't customize the text much.

### Describe the solution you'd like

Example, this will in a markdown card show colors based in state, in  a mushroom template card it will just show html/markdown as text.

`{% if is_state("input_boolean.guest_home", "on") -%}
<font size="+1"> Guest mode:<font color="lime"> On</font></font>
{%- endif %}{% if is_state("input_boolean.guest_home", "off") -%}
<font size="+1">Guest mode:<font color="teal"> Off</font></font> 
{%- endif %}`

piitaya:
There is no way add custom style in text to keep design clean and uniform across cards 😅

Muenchen-Michi:
HTML inside Template-Card would be great.
I want to have 2 information within the secondary info. A state and a power. The state on the left, the power aligned to the right.
My idea was to insert a <span> that I can pull right using css.
Without HTML it is not even possible to insert protected spaces to create a larger space between 2 texts.

sibbl:
> There is no way add custom style in text to keep design clean and uniform across cards 😅

To be honest I think by just allowing to make text e.g. bold or italic it can still be clean and uniform.

Since the template card is one of the most used cards of the project, maybe this could be reevaluated.

Together with card_mod it would allow some very complex scenarios if we could just allow specific parts of the template to be wrapped even just in `<b>` or `<i>` tags to mention the most basic elements.

Even Android and iOS notifications, which focus on clean design and uniformity, allow some very basic styling possibilities, see the [HA companion docs](https://companion.home-assistant.io/docs/notifications/notifications-basic/#notification-message-html-formatting).

PR #1397, opened 2024-03-29
Allow use of home-assistant icons in template card secondary.

### LINKED ISSUE #327: Display icons in text section
I would to display icon in text section, instead of displaying status

Maybe we could use it to display icon:
```
{ icon:'mdi:sofa' }
```

**Additional context**
image

thybag:
## Description

This PR allows mushroom card's to use home-assistant icons in their secondary state area by defining these as `:mdi:home-assistant:` within the templates.

This allows template cards for example to be setup to show an icons for any device that is currently turned on, e.g.
image 
image

This can also be used to allow for inline icons, for example next to displayed temperature values such as in the link ticket.

This has been implemented via adding a `parseIcons` function to the `state-info.ts`, which uses a rejex to swap out bracketed values, e.g. `:mdi:icon-name:` for `<ha-icon icon="mdi:icon-name"></ha-icon>`. I've wrapped this in a fragment in order to render the added HTML as part of the TemplateResult. I've additionally tweaked the css to ensure the icons are an appropriate size for the context.

I've never worked with TypeScript or Lit components before, so apologies if I've set them up wrong/butchered them a bit.  Please let me know if there are cleaner ways of approaching this.

I'd be appreciative of any guidance as to whether this change would warrant a documentation update, assuming this functionality is something that is seen as desirable to have added.

Re: code style - I ran the prettier command to format my change. I wasn't able to easily locate any additional document on the code style choices, beyond attempting to be similar to what is already there.

Note:
* I've currently implemented icons as `{mdi:sofa}` as its nice and simple, although it'd be easy to swap the regex to look for `{ icon:'mdi:sofa' }` instead.
* The change will technically allow raw html rendered in the secondary (as it bypasses the escaping in order to add the icons). Given this content is edited by the owner, I don't think* this should be a problem (this is also possible via the HA markdown cards) but is probably worth flagging in case this is a concern .

## Related Issue

This PR fixes or closes issue: fixes https://github.com/piitaya/lovelace-mushroom/issues/327

## Motivation and Context

I currently use template cards for each room of my house, which then contain logic to show a Unicode icon for each device that's is turned on in the given room within the cards secondary area. Unicode is far more limited in the range of icons i can use, so I had a go at this in order to allow for much better identification of specific devices - ie. is it a lamp or main light that's on.

Other use cases include in-line use of icons next to figures displayed (power usage by a device, a temperature reading etc)

## How Has This Been Tested

I've primarily tested this on my own home assistant instances by swapping out the build JS imported by hacs. 
I've tested a variety of icons, although as far as I'm aware only the template card allows me to directly control what text (and thus icons) are output in the secondary area.

I could make the behaviour available to `content` on the chip templates or names, if that was a desirable extension to this.

Tested in
* Chrome 123.0.6312.86 (windows)
* Edge 124.0.2478.10 (windows)
* Firefox 124.0.1 (windows)

## Types of changes

## Checklist

### DIFF
--- a/src/cards/template-card/template-card.ts
+++ b/src/cards/template-card/template-card.ts
@@ -153,24 +153,21 @@ export class TemplateCard extends MushroomBaseElement implements LovelaceCard {
             return undefined;
         }
 
-        // Get value as a string to enable string operations
-        const textAsString = `${text}`;
-
         // Convert :mdi:abc: format strings to HA icons.
-        const parsedString = textAsString.replace(
+        const parsedText = text.replace(
             /:([a-z]{3}\:[a-z0-9\-]+):/g,
             '<ha-icon icon="$1"></ha-icon>'
         );
 
         // If initially recorded length & current length match then we have not
         // parsed any icons
-        if (parsedString.length === textAsString.length) {
-            return html`${parsedString}`;
+        if (parsedText.length === text.length) {
+            return html`${parsedText}`;
         }
 
         // Otherwise, some icons must have been found. Convert to fragment to avoid icon
         // markup being escaped
-        const fragment = document.createRange().createContextualFragment(parsedString);
+        const fragment = document.createRange().createContextualFragment(parsedText);
         // Return content
         return html`${fragment}`;
     }
--- a/src/cards/template-card/template-card.ts
+++ b/src/cards/template-card/template-card.ts
@@ -147,6 +147,34 @@ export class TemplateCard extends MushroomBaseElement implements LovelaceCard {
             : this._config?.[key];
     }
 
+    protected parseIcons(text: string | undefined): TemplateResult | undefined {
+        // Skip logic if no secondary is defined
+        if (text === undefined) {
+            return undefined;
+        }
+
+        // Get value as a string to enable string operations
+        const textAsString = `${text}`;
+
+        // Convert :mdi:abc: format strings to HA icons.
+        const parsedString = textAsString.replace(
+            /:([a-z]{3}\:[a-z0-9\-]+):/g,
+            '<ha-icon icon="$1"></ha-icon>'
+        );
+
+        // If initially recorded length & current length match then we have not
+        // parsed any icons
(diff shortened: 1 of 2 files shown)


piitaya:
Hello 👋
If this is used for template card, the parsing should be located in template card because state-info component is using by every card.

Also, for the icon format with brackets, I'm not sure because bracket already have a role in jinja template. Github is using `:emoji:` format for emoji. I wonder if we can have something like `:mdi:home:`. What do you think?

thybag:
Hi @piitaya 

Thank you for the quick feedback!

I've reworked the logic as you suggested so that
- The parse icon functionality now lives in the template card itself rather than state-info
- The icon format is now `:mdi:icon-name:`
- I've added a check to avoid the `createContextualFragment` if the secondary doesn't contain any matched icons

I've also retested on my local and everything still appears to work happily for me.

Re: the icon format, I think the only thing that feels a bit off with `:mdi:home:` is that it ends up with 3 `:` - although I think that's unavoidable assuming support for non-mdi icons is wanted and isn't really an issue in normal use. I guess the other aspect worth considering re: the original tickets proposal is that if you went with something like `{ icon:'mdi:sofa' }`, it'd be a straightforward evolution if you ever wanted to support color etc `{icon:"mdi:sofa" color:"#fff"}`. 
That said, I do agree the `{` approaches risk it being confused for the jinja templates.

TurnrDev:
If you added markdown support, we could use the `ha-icon` attribute
`<ha-icon icon="mdi:temperature-celsius"></ha-icon>`

thybag:
Hmm, markdown support could be a more flexible approach.

I see similar suggestions ages back https://github.com/piitaya/lovelace-mushroom/issues/767 and https://github.com/piitaya/lovelace-mushroom/issues/1189 more recently.

VincentKueszter:
@thybag Any news on this? This would be a great feature.

kattcrazy:
Ooh this would be cool to have!

PR #1420, opened 2024-05-07
Remove hide_icon from showcase

martindybal:
## Description
This just fixing UI editor in development `template-view.yaml`

`hide_icon` has been removed from all cards in v2.0.0.

At path: hide_icon -- Expected a value of type `never`, but received: `true`


piitaya (approved):

piitaya:
Thank you!

Issue #1421, opened 2024-05-07
template-cards custom buttons


martindybal:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

I was limited by the template card in some scenarios. I tried to add frequent actions to the device, but I couldn't find a way to do it. 

### Describe the solution you'd like

Add custom buttons for template card. To allow frequent actions to be used from dashboard. Like this 
![image](https://github.com/piitaya/lovelace-mushroom/assets/7644099/d74aec9d-a80a-428e-a8c9-23048de104cc)

martindybal:
This is duplication of https://github.com/piitaya/lovelace-mushroom/issues/887

Issue #887, opened 2022-12-08
[Feature]: Allow buttons in Template card


netsoft-ruidias:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

I would like the possibility to add buttons with custom actions in `mushroom-template-card`.

When we create a template card, we can customize all properties to have the look and feel of the other cards, but with custom data. It would be interesting to also be able to define actions.

### Describe the solution you'd like

The solution should be something along the line of defining the action through yaml, like this:

```yaml
- type: custom:mushroom-template-card
  entity: sensor.xyz
  primary: '{{ state_attr(entity, ''friendly_name'') }}'
  secondary: '{{ states(entity) }}' XY
  icon: |
    {% set target=state_attr(entity, 'temperature') %}
    {% set current=state_attr(entity, 'current_temperature') %}
    {% if current < target %}
      mdi:thermometer-chevron-up
    {% else %}
      mdi:thermometer-chevron-down
    {% endif %}  
  icon_color: |-
    {% set state=states(entity) %}
    {% if state=='heat_cool' %}
      var(--rgb-orange)
    {% elif state=='heat' %}
      var(--rgb-red)
    {% else %}
      var(--rgb-disabled)
    {% endif %}
  tap_action:
    action: call-service
      service: browser_mod.more_info
      service_data:
        entity_id: sensor.xyz
        deviceid:
          - this
  # new feature starts here:
  actions:
    - type: action
      icon: mdi:fan-speed-1
      tap_action:
        action: call-service
        service: fan.set_fan_speed
        service_data:
          entity_id: entity.xxx
          fan_speed: Silent
    - type: action
      icon: mdi:fan-speed-2
      tap_action:
        action: call-service
        service: fan.set_fan_speed
        service_data:
          entity_id: entity.xxx
          fan_speed: Standard
    - type: action
      icon: mdi:fan-speed-3
      tap_action:
        action: call-service
        service: fan.set_fan_speed
        service_data:
          entity_id: entity.xxx
          fan_speed: Strong
    - type: conditional
      conditions:
        - entity: binary_sensor.xxxyyyzzz
          state: 'on'
      action:
        icon: mdi:weather-windy
        tap_action:
          action: call-service
          service: fan.set_fan_speed
          service_data:
            entity_id: entity.xxx
            fan_speed: Turbo
```

[...]

Issue #1259, opened 2023-09-28
[Feature]: Mushroom version of the tile card


Silther:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Is your feature request related to a problem?

no

### Describe the solution you'd like

I would like to see a mushroom version of the in built tile card, as both look modern and minimalistic, but are different enough in design so that they cannot be easily combined in the same dashboard.
Also mushroom cards are way more customizable (hide name icon etc.)

![image](https://github.com/piitaya/lovelace-mushroom/assets/97352673/2b38b776-fbbc-46d2-9f9a-4594abb0928e)

### Describe alternatives you've considered

Change the mushroom design to look more like the (hopefully uniform) modern Home Assistant card design.

piitaya:
Hi 👋 I agree, some changes about margin or font-size should be done.

Basically, tile card is an improvement of Mushroom card in term of accessibility support and flexibility with features. What mushroom features are you missing in tile card?

BBESINET:
+1 with the above statement.

For me two other things are missing in the Tile cards compare to mushroom:
- There is no template in tile cards. 
- There are 3 option for action (single press, double press and long press) compared to only two with tiles (single press on card, single press on icon) => I like to "secure" some action with a long press.

Once you have populated your dashboard with mushroom template card, you are trapped with mushrooms (but for the best!)

Phontana:
@piitaya I would like to see that the mushroom cards have the same styling as the native tile cards, for example the secondary text font style is different. Currently I use card_mod to set it to primary text color, just like the tile card. This should not be necessary.

Mariusthvdb:
> Hi 👋 I agree, some changes about margin or font-size should be done.
> 
> Basically, tile card is an improvement of Mushroom card in term of accessibility support and flexibility with features. What mushroom features are you missing in tile card?

most notably might be the horizontal-layout with all elements in the single height:

<img width="507" alt="Scherm­afbeelding 2024-01-31 om 11 49 12" src="https://github.com/piitaya/lovelace-mushroom/assets/33354141/b7377796-bcdc-4786-b8ce-fb020a821b19">

and the hide element when off

<img width="583" alt="Scherm­afbeelding 2024-01-31 om 11 50 22" src="https://github.com/piitaya/lovelace-mushroom/assets/33354141/15ca19a5-e12e-4dcc-9105-bd7b30743809">

[...]

PR #1422, opened 2024-05-07
Add custom buttons to template-card

martindybal:
## Related Issue

This PR fixes or closes issue: fixes #
https://github.com/piitaya/lovelace-mushroom/issues/887
https://github.com/piitaya/lovelace-mushroom/issues/1421
## Motivation and Context
As I mentioned in the last issue, I was a bit limited by the template card in some scenarios. It would be really great if I could have a quick action directly on my dashboard.

## How Has This Been Tested
I tested it locally on my HA instance. I also created a sample in the dev box at `/lovelace-mushroom-showcase/5`.

image

## Types of changes

## Checklist

### DIFF
--- a/.hass_dev/views/template-view.yaml
+++ b/.hass_dev/views/template-view.yaml
@@ -58,6 +58,67 @@ cards:
     columns: 2
     square: false
   - type: vertical-stack
+    title: Custom buttons
+    cards:
+      - type: grid
+        columns: 2
+        square: false
+        cards:
+          - type: custom:mushroom-template-card
+            primary: Hello, {{user}}
+            secondary: How are you?
+            icon: mdi:home
+            first_button_icon: mdi:arrow-up
+            first_button_tap_action:
+              action: none
+            second_button_icon: mdi:arrow-right
+            second_button_tap_action:
+              action: none
+            third_button_icon: mdi:arrow-down
+            third_button_tap_action:
+              action: none
+            fourth_button_icon: mdi:arrow-left
+            fourth_button_tap_action:
+              action: none
+      - type: grid
+        columns: 2
+        square: false
+        cards:
+          - type: custom:mushroom-template-card
+            primary: Hello, {{user}}
+            secondary: How are you?
+            icon: mdi:home
+            layout: vertical
+            first_button_icon: mdi:arrow-up
+            first_button_tap_action:
+              action: none
+            second_button_icon: mdi:arrow-right
+            second_button_tap_action:
+              action: none
+            third_button_icon: mdi:arrow-down
+            third_button_tap_action:
+              action: none
+            fourth_button_icon: mdi:arrow-left
+            fourth_button_tap_action:
+              action: none
+      - type: custom:mushroom-template-card
+        primary: Hello, {{user}}
+        secondary: How are you?
+        icon: mdi:home
+        layout: horizontal
+        first_button_icon: mdi:arrow-up
+        first_button_tap_action:
+          action: none
+        second_button_icon: mdi:arrow-right
+        second_button_tap_action:
+          action: none
+        third_button_icon: mdi:arrow-down
+        third_button_tap_action:
+          action: none
+        fourth_button_icon: mdi:arrow-left
+        fourth_button_tap_action:
+          action: none
+  - type: vertical-stack
     title: Layout
     cards:
       - type: grid
--- a/docs/cards/template.md
+++ b/docs/cards/template.md
@@ -11,22 +11,30 @@ A template card allows you to build a custom card. You can use `entity` as a var
 
 All the options are available in the lovelace editor but you can use `yaml` if you want.
 
-| Name                  | Type            | Default     | Description                                                                                                                         |
-| :-------------------- | :-------------- | :---------- | :---------------------------------------------------------------------------------------------------------------------------------- |
-| `icon`                | string          | Optional    | Icon to render. May contain [templates](https://www.home-assistant.io/docs/configuration/templating/) \*.                           |
-| `icon_color`          | string          | Optional    | Icon color to render. May contain [templates](https://www.home-assistant.io/docs/configuration/templating/).                        |
-| `primary`             | string          | Optional    | Primary info to render. May contain [templates](https://www.home-assistant.io/docs/configuration/templating/).                      |
-| `secondary`           | string          | Optional    | Secondary info to render. May contain [templates](https://www.home-assistant.io/docs/configuration/templating/).                    |
-| `badge_icon`          | string          | Optional    | Badge icon to render. May contain [templates](https://www.home-assistant.io/docs/configuration/templating/).                        |
-| `badge_color`         | string          | Optional    | Badge icon color to render. May contain [templates](https://www.home-assistant.io/docs/configuration/templating/).                  |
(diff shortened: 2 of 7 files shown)


BBE-FR:
For information, up to what level are the button customisable ?
For example can we like with a template chip:
- define dynamic icon/color ?
-  define tap, hold and double tap actions ?
If yes that is exactly what I wanted to create a room card ! 

Whatever the customization, Many Thanks @martindybal and I hope it goes up to release !

martindybal:
@BBE-FR Thank you very much for your feedback and support!

Regarding your questions:

Dynamic icon/color: This is not supported in the current PR, but can be easily added if needed.
Tap, hold and double tap actions: Only tap actions are currently supported, but again, they can be implemented if desired.

Before I proceed with these additions, I need to hear from @piitaya if he wants to finish this PR.

Thanks for your suggestions.

piitaya:
Hi 🙂
I like the idea but if we implement this, I think it must be an array instead of 4 elements. 

Also I have concerns to bring new features like this to Mushroom because the gap between Mushroom and tile card become less important with HA updates. Many features request (icon tap action, card features, inverted cover slider, etc...) to Mushroom are improvements done to the tile card so it become hard to do the work in double.

I have some options but I didn't decided about the best solution : 
1. Implement all Mushroom features to tile card and deprecate Mushroom usage
2. Implement all Mushroom features to tile card except template features. Mushroom will become a kind of "advanced tile card" with template support and advanced features.
3. Freeze Mushroom dev and maintain it to avoid regression with HA updates.

The problem with adding this option is that it will encourage people to not use tile card and it can lead to more breaking changes in the future if we want to fully merge Mushroom to Home Assistant.

BBE-FR:
@piitaya I imagine that the question is hard...

Option 1 would be Nice...

The lack of template card and chips are the only real things that make me stay with mushroom... without template its difficult to go back to tiles...

For the rest tiles are sligtly less customisable but its enought for 99% of time...

jhemak:
+1 for Option 1 ... I would love to switch over to Tile Card 100% but am a heavy user of template and chip.

martindybal:
@piitaya I fully understand. The reason for 4 elements is layout, but it's not a problem to use an array with limited size. I'm just not sure about the UI editor. 

I have no problem helping with the development of the Tile Card. I've switched to it myself in a few scenarios with mushrooms. My main reason for sticking with Mushrooms is 
1. Horizontal layout
  - I added card_mod
2. Better tap and hold action management
  - hold_action doesn't work above the icon. It uses icon_tap_action even with hold. 
3. Features like animated icons 

I'm using combination of Mushrooms and tile card for all my room like this
image

BBE-FR:
@piitaya 
Do we have any chance to have templates in tile cards the same way as in mushroom ( template defining icon, color, primary and secondary information) ? 

**If not planed** it might be worth to release this feature, I think there is a long lasting request for this kind of function in the mushroom Community. Y think we are numerous to use stack in to fuse template cards (or other mushroom when template is not mandatory) with chips to create additional buttons.
Like in my personal exemple there: https://github.com/piitaya/lovelace-mushroom/issues/55#issuecomment-2046983165

**If planned**, there is an HACS project to allow to use controls: https://github.com/Nerwyn/service-call-tile-feature this could be embeded into the "template tile" to allow to define buttons into this future "_advanced templatable tile_"...

BBE-FR:
> @BBE-FR Thank you very much for your feedback and support!
> 
> Regarding your questions:
> 
> Dynamic icon/color: This is not supported in the current PR, but can be easily added if needed. Tap, hold and double tap actions: Only tap actions are currently supported, but again, they can be implemented if desired.
> 
> Before I proceed with these additions, I need to hear from @piitaya if he wants to finish this PR.
> 
> Thanks for your suggestions.

I used chips instead of button with stack in in my cards in order to keep the mushroom look and to have the ability to define the button through graphical interface (including full control of the icons/colors through the template chips)...
Description can be found here: https://github.com/piitaya/lovelace-mushroom/issues/55#issuecomment-2046983165

@martindybal Your proposal is already a great improvement and I fully support this. If at the end you manage to have the same capacity than chips, that would be 100% marvelous to me !

martindybal:
@BBE-FR I'm waiting for a reply from @piitaya

martindybal:
@piitaya can you please comment? Does it make sense to continue on this or would TileCard be the future?

piitaya:
Sorry for the late reply, I was busy these days. 
I'm not sure tile card will have templating features because it's too complex for most users and it can lead to performance issue if there are many templates.
However, I do agree that Mushroom should become an "advanced tile card" with many features, And of course, these features should be edited with a GUI.

For now, I would say we close this PR and I will work on a "advanced" version of the card but I have no deadline to share... However, I personally only use Mushroom card for the template card so I would love to move forward 😅

BBE-FR:
> I will work on a "advanced" version of the card but I have no deadline to share... However, I personally only use Mushroom card for the template card so I would love to move forward

This might be a way to advance on the subject:
instead of updating endlessly mushrooms or risking to have to much breaking changes as stated here: https://github.com/piitaya/lovelace-mushroom/issues/1259#issuecomment-2325865362

One option could be:
- to create a new "advanced tile" custom card in HACS (possibly called called toadstool ??) that would basically be a tile card structure and look, but with full control on icon and templating. 
- to let the mushroom as it is, without new functionalities, and only maintain it, as long as feasible.
- This card (or card family) could then, evolve to implement new functionalities (like buttons, sliders, etc, with same design as the tile card, but fully controllable by the user).

I think that otherwise, there will always be pros and cons to stay with mushroom "classical" cards and some will never migrate to tiles.

My personal example is as follow: 
I am willing to go full Tiles (despite I prefer the interface of mushroom with collapsible controls and selectors) but restrained by the lack of:
- template card (more than 75% of my mushroom cards)
- chips (that can be placed anywhere and not just "on top")

With the latest badges (especially if they can be placed on top of sections) I could get rid of chips and use only Tiles. The last real functional blocking point for me is templating capability in Tile or Tile-like card...

j9brown:
I'm debating cherry-picking this PR into my own branch of Mushroom because it solves an immediate problem I have with wanting to add buttons to end of template cards with a horizontal appearance.  Tile doesn't support a horizontal appearance currently, so I'm pretty much stuck using Mushroom.

But if this is the way the wind is blowing, then maybe I should redo my layouts to not rely on the horizontal appearance.

To be honest, I think that it would be better to expand the built-in Tile card's feature set to cover the needs of power users too rather than to make a separate "advanced Tile" card that's sort of the same and sort of different and sort of awkward to find and with significant duplication of development effort overall.  Expanding the built-in Tile card (or making more flavors of built-in Tile cards) would help users progressively discover the capabilities they need as they need them.  It also helps the Home Assistant user base as a whole converge on more of the same design vocabulary.  Some programs address this progressive discovery issue by hiding more complex or less frequently used features behind toggles such that a hard bifurcation of the UI isn't necessary.

martindybal:
@j9brown I do exactly that. I moved whole dashboard to TileCard with service call tile customer feature. I’m using card_mod for horizontal layout. 
image

piitaya:
Unfortunately, I'm closing this PR as I'm planning to release a similar feature in the following months with integration with tile card and card features.

PR #1726, opened 2025-06-13
feat(lock): add config option to show open door button when locked

### LINKED ISSUE #876: [Bug]: Inconsistent behaviour of Lock card: the Open button shows randomly
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Current Behavior

The Lock Card should show the Open button based on the supported_features attribute of the lock, but since I migrated to Mushroom, I don't always see the second button on the card, it appears randomly.

I'll show 2 screenshots where you can see the same card with only 1 button and with the 2nd, the lock is always the same (Nuki). You can also see the attribute of the lock: `supported_features: 1`.

image
image

### Expected Behavior

_No response_

### Steps To Reproduce

_No response_

### Context

image

### Environment

```markdown
- Browser: Chrome
- Home Assistant 2022.11.4 / Frontend 20221108.0
- Mushroom: v2.4.1
```

### Anything else?

_No response_

kvanzuijlen:
## Description
This PR allows showing the "Open door" button even when the lock is locked. When the lock card was initially implemented, not allowing this seemed to have been a conscious choice. But since there are a lot of locks that do support opening the door from a locked state, it makes sense to add this.

image

## Related Issue

This PR fixes or closes issue: closes #876

## Motivation and Context
This PR implements the requested feature in #876. By making this an option, users with locks that don't support opening from the locked state can choose not to enable this.

As the `lock.open` action might leave the door ajar, some users might also find this button undesirable, as it can't be relocked without manual intervention / a (smart) door closer.

## How Has This Been Tested
Tested it on my own HA instance by uploading to the /config/www folder and loading the resource instead of the HACS resource. The attached screenshot is from my own instance. Disabling the switch will hide the "door open" button when locked, and it will still show (if supported) when the door is unlocked.

## Types of changes

-   [x] 🌎 Translation (addition or update a translation)
-   [x] ⚙️ Tech (code style improvement, performance improvement or dependencies bump)
-   [x] 📚 Documentation (fix or addition in the documentation)

## Checklist

### DIFF
--- a/docs/cards/lock.md
+++ b/docs/cards/lock.md
@@ -11,19 +11,20 @@ A lock card allows you to control a lock entity.
 
 All the options are available in the lovelace editor but you can use `yaml` if you want.
 
-| Name                | Type                                                | Default     | Description                                                                         |
-| :------------------ | :-------------------------------------------------- | :---------- | :---------------------------------------------------------------------------------- |
-| `entity`            | string                                              | Required    | Lock entity                                                                         |
-| `icon`              | string                                              | Optional    | Custom icon                                                                         |
-| `name`              | string                                              | Optional    | Custom name                                                                         |
-| `layout`            | string                                              | Optional    | Layout of the card. Vertical, horizontal and default layout are supported           |
-| `fill_container`    | boolean                                             | `false`     | Fill container or not. Useful when card is in a grid, vertical or horizontal layout |
-| `primary_info`      | `name` `state` `last-changed` `last-updated` `none` | `name`      | Info to show as primary info                                                        |
-| `secondary_info`    | `name` `state` `last-changed` `last-updated` `none` | `state`     | Info to show as secondary info                                                      |
-| `icon_type`         | `icon` `entity-picture` `none`                      | `icon`      | Type of icon to display                                                             |
-| `tap_action`        | action                                              | `more-info` | Home assistant action to perform on tap                                             |
-| `hold_action`       | action                                              | `more-info` | Home assistant action to perform on hold                                            |
-| `double_tap_action` | action                                              | `more-info` | Home assistant action to perform on double_tap                                      |
+| Name                           | Type                                                | Default     | Description                                                                                           |
+| :----------------------------- | :-------------------------------------------------- | :---------- | :---------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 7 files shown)


kvanzuijlen:
@piitaya Is there anything you need from me to get this merged? Thanks for all the great work on these cards!

davidmateos:
+1 to this feature!

piitaya (commented):
Hi 👋 
Thank you for your contribution. Sorry for the late review, I was busy during the last weeks.

In HA more info, the button is always here but with a confirmation. I wonder if we should use the same pattern in the lock card instead of adding an option. What do you think?

https://github.com/user-attachments/assets/aff8c176-6542-4ef7-a53e-429dbc8ab4f9

kvanzuijlen:
Hi @piitaya,

No worries, thank you for all the time and effort you put into this awesome repository!

> In HA more info, the button is always here but with a confirmation. I wonder if we should use the same pattern in the lock card instead of adding an option. What do you think?

I personally think it would be fine, but I can also imagine users wanting to keep the current style where the button is only visible when unlocked. If you do think it should always be shown if supported, I will make the change.

I also wanted to discuss the placement of the config option (if you decide to keep it). Should it be between Icon Type and Tap Behavior, or is the placement fine as-is?

EDIT: How would you like to see the confirmation happen? I think the button will be too small for text.

Thanks!

PulsarFX:
> Hi 👋 Thank you for your contribution. Sorry for the late review, I was busy during the last weeks.
> 
> In HA more info, the button is always here but with a confirmation. I wonder if we should use the same pattern in the lock card instead of adding an option. What do you think?


I absolutely fafour a 2 step unlatch operation! This prevents accidentally opening your front door, i.e. when you are not at home and it would be the worst case to do so. 
I can't understand why the card has a single tap to unlatch, but the details view has the far more safe double tap to unlatch. 
Double tap should be absolutely the default on any unlatch button!
Imagine scrolling your HA view on a bumpy ride and your phone thinks "ah user tapped the unlatch button", and bam your front door springs open. Completely crazy.

distante:
+1 For the 2 steps operation! I was going to do a Feature request for that :)

Issue #1022, opened 2023-02-12
[Bug]: Climate card showing temperature control for heat/cold/off instead of just the current mode


clowncracker:
### Requirements

- [X] I have updated Mushroom to the latest available version
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Current Behavior

Previously the climate card would only show temperature control for the current mode the thermostat was in. Now it's showing controls for modes that aren't being used, this adds a lot of clutter to the card.

### Expected Behavior

I expect the temperature control to be accessible to the current mode the thermostat is on.

### Steps To Reproduce

The climate card now has three temperature controls instead of one. This was not an issue in 2.4.1.  I updated from 2023.2.0 to 2023.2.1 and that fixed the issue, not sure why.  I am now on 2023.2.3 and the issue seems to be back.

### Context

YAML state
```yaml

```

### Environment

```markdown
- Browser: Firefox
- HA Version: 2023.2.0 
- Mushroom: 2.5.3
```

### Anything else?

![Thermostat](https://user-images.githubusercontent.com/49218426/218326528-0265a77c-2669-4e70-8ef5-235b5c6c7225.jpg)

cromelex:
My thermostat supported by zigbee2mqtt was updated and it now announces a "cooling setpoint".
Now, this means Mushroom Climate card also shows the cooling temperature, even when set to OFF or HEAT.

This adds a lot of clutter.

Happening on latest.

In my opinion of the HVAC mode is set to HEAT, the card should not show the cooling setpoint, and vice versa.

![image](https://github.com/user-attachments/assets/878dc915-6298-40e4-96b2-fcccf33c0f74)


Could we get a fix or enhancement?

1iverea9er:
Same problem.

How it looks compared to the built-in `tile` card. `heat` mode:
<img width="765" height="448" alt="Image" src="https://github.com/user-attachments/assets/e274d45d-4112-4fff-b06e-61525a8a1be2" />
`heat_cool` mode:
<img width="767" height="446" alt="Image" src="https://github.com/user-attachments/assets/e8bd3193-eb02-452a-a47e-bbfc98c2dc3c" />

The card displays the target temperature settings correctly in the `heat_cool` mode, but in the `heat` or `cool` modes, it displays 3 target temperature control fields. They seem to correspond to the parameters `temperature`, `target_temp_low`, and `target_temp_high`

- Browser: Chrome
- HA Version: 2025.10.0 
[...]

PR #1816, opened 2025-10-16
Fix climate temp control

1iverea9er:
## Description

Fixes incorrect HVAC state handling in the climate temperature control card in `heat` or `cool` mode for Dual-point Thermostat like [this](https://esphome.io/components/climate/thermostat/).

## Related Issue

This PR fixes or closes issue: fixes [#1022](https://github.com/piitaya/lovelace-mushroom/issues/1022)

## How Has This Been Tested

Verified via console logs and UI behavior

## Types of changes

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ⚙️ Tech (code style improvement, performance improvement or dependencies bump)
-   [ ] 📚 Documentation (fix or addition in the documentation)

## Checklist

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,7 +80,6 @@
       "integrity": "sha512-2BCOP7TN8M+gVDj7/ht3hsaO/B/n5oDbiAyyvnRlNOs+u1o+JWNYTQrmpuNp1/Wq2gcFrI01JAW+paEKDMx/CA==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.3",
@@ -3658,7 +3657,6 @@
       "integrity": "sha512-5gTmgEY/sqK6gFXLIsQNH19lWb4ebPDLA4SdLP7dsWkIXHWlG66oPuVvXSGFPppYZz8ZDZq0dYYrbHfBCVUb1Q==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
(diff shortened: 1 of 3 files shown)


1iverea9er:
Hello, @piitaya . 
Please correct me if I am wrong. 
However, it appears to me that `supported_features` is a static parameter of the climate entity. That is, it does not depend on state (climate mode) in any way. Therefore, it is not suitable for us as the main trigger for determining the single/range target temperature.

piitaya:
Exactly, it is a static value. Then the integration has to define `temperature` (for single target) or `target_temp_low` and `target_temp_high` (for range).
It should be defined like that : https://github.com/home-assistant/frontend/blob/dev/src/state-control/climate/ha-state-control-climate-temperature.ts#L394C3-L411C1

1iverea9er:
What I found out.

[Here](https://github.com/home-assistant/frontend/blob/dev/src/state-control/climate/ha-state-control-climate-temperature.ts), the renderer uses checks of the following type:
`if () ... elseif() ... endif`
Mushroom uses two separate blocks like:
` if () ... endif`
Therefore, in the latter case, if the climate entity supports both "single" and "range" and the corresponding attributes have numeric values, all three fields are displayed. This is what happens in `cool` and `heat` modes.
But the trick is that in `heat_cool` mode, the `temperature` attribute is `null`.
Therefore, to solve the problem, we actually only need to set the priority of "single" over "range" when checking conditions. That is, if `temperature` is not null, we display one field with "single" target temperature regardless of the results of the "range" checks. This is how tile card work.

However, I added a check for `supported_features`.

Please take a look at the latest commit.

Best Regards, Gene

1iverea9er:
@piitaya Hi, could you take a look at the latest commit? To be honest, I thought I had made all the necessary changes. These changes have been tested and are working. Support for `supported_features` has been added in the commit, although this is not the cause of the bug.

I see a changes requested review, but I don’t see any comments.
Could you please clarify what needs to be changed?

piitaya (approved):
Hi 👋

Sorry for the late review. The changes are ok and in sync with HA front-end.

PR #1839, opened 2026-01-01
Improve color contrast handling in light chip and card components

### LINKED ISSUE #1838: [Bug]: Light card goes gray when color set to yellow
### Requirements

- [x] I checked the troubleshooting section in the README to verify that I have the latest Mushroom version.
- [x] I did a search to see if there is a similar issue or if a pull request is open.
- [x] I have read the state of card mod support (in pin issues) and understand card mod support is not provided in this repository.

### Current Behavior

When using the "Use light color" setting with the light card, the icon and brightness slider is being set to gray when the light colour is set to a yellow (in this case #fff700) in the color slider. This only occurs when running a light theme.

Image

### Expected Behavior

#fff700 is easily visible when displayed on a white background

Image

### Steps To Reproduce

Set Home Assistant theme to light mode
Create a card `type: custom:mushroom-light-card` with `use_light_color: true`
Set light colour to #fff700 (or similar)

### Environment

```markdown
- Browser: Chrome
- HA Version: 2025.12.4
- Mushroom: 5.0.8
```

### Anything else?

Currently the below functions are only checking for lightness:

https://github.com/piitaya/lovelace-mushroom/blob/64775d6ad79cf2b254b02889ecf555c70adb5e27/src/cards/light-card/utils.ts#L27-L46

This should also check chroma to filter only the colors closer to white. Happy to test some options and open a PR

wiiaam:
## Description

Updated `light-card` and `light-chip` cards to apply contrast corrections for light colors instead of reverting to default/theme colors.

Lightness checks and variations are now done through HSV, instead of LCH. This matches the behaviour of the [HA default brightness slider](https://github.com/home-assistant/frontend/blob/dev/src/state-control/light/ha-state-control-light-brightness.ts#L50-L57)

As previous, contrast correction is only applied when the theme is not set to dark.

Example - light set to #FFFDED
image

## Related Issue

This PR fixes or closes issue: fixes #1838

## Motivation and Context

The `light-card` and `light-chip` cards are currently using lightness (LCH) only to determine if they are too light to be displayed. As this doesnt account for chroma, some shades of yellow are being incorrectly flagged as too light. 

The simple fix would be to add an additional chroma check to the `isColorLight` and `isColorSuperLight` functions, however removing these functions entirely and changing to a contract correction improves visual feedback by keeping the UI consistent with the selected light color.

## How Has This Been Tested

Tested on HA dev with various themes.

## Types of changes

-   [x] 🐛 Bug fix (non-breaking change which fixes an issue)
-   [ ] ⚙️ Tech (code style improvement, performance improvement or dependencies bump)
-   [ ] 📚 Documentation (fix or addition in the documentation)

## Checklist


piitaya (commented):
Hi 👋
I think it's a great idea to use the same logic as HA for this color.

Can you update the code to use the helpers so we can use similar code than HA codebase?

I think it's ok to also use this in dark mode right? So it's consistent with HA.

```ts
const hsvColor = rgb2hsv(entity.attributes.rgb_color);

// Modify the real rgb color for better contrast
if (hsvColor[1] < 0.4) {
  // Special case for very light color (e.g: white)
  if (hsvColor[1] < 0.1) {
    hsvColor[2] = 225;
  } else {
    hsvColor[1] = 0.4;
  }
}
return rgb2hex(hsv2rgb(hsvColor));
```

wiiaam:
Have updated to use the same helpers, and have removed the light mode check so it applies it regardless of theme. Also found a stray console.log when testing which I've removed.

This conversion is technically not needed on dark mode, since the contrast correction only occurs when the color is near white. However I do agree that matching the current HA logic makes sense for consistency, especially since a lot of people will be mixing mushroom and tile cards due to their similarity.

The only issue I can see with dark mode is if the HSV Value was nearing 0. This usually doesnt happen since the HA color wheel doesnt allow changes in Value, however I can manually change Value in something like WLED, and it will produce this - 

image


`if (hsvColor[1] < 0.1) hsvColor[2] = 225;` prevents this from happening when the color is near white, but I do think theres room to improve on this. At this point I might consider opening a PR on frontend to move the contrast correction to the common library, then it can be vendored back here.

wiiaam:
Hey @piitaya could I possibly get another look at this? Please let me know if theres anything else that needs a tweak.

Understand you're mainly busy with HA frontend. Cheers

piitaya (approved):
Sorry for the late review. The PR is ok to be merged ! Thank you !

PR #1917, opened 2026-06-26
Fix documented default for climate card tap_action

### LINKED ISSUE #1245: [Bug]: <title>
### Requirements

- [X] I checked the troubleshooting section in the README to verify that I have the latest Mushroom version.
- [X] I did a search to see if there is a similar issue or if a pull request is open.

### Current Behavior

Documentation for the climate card says the default value for tap_action is more-info but it's actually toggle.

### Expected Behavior

Default value for tap_action should be documented as toggle instead of more-info

### Steps To Reproduce

See documentation

### Context

_No response_

### Environment

_No response_

### Anything else?

_No response_

cycsmail:
## Description

The climate card docs list `tap_action`'s default as `more-info`, but the card actually defaults to `toggle`. Updates the table row to match the source.

## Related Issue

Fixes #1245

## Motivation and Context

Documented default didn't match the actual behavior, so anyone reading the docs to configure the card would get the wrong value. Verified against the card source (defaults to `toggle`) and ran prettier, which is clean. Docs only.


piitaya (approved):
Thank you !

